PR #4, opened 2024-03-24
Some properties are parsed but never assigned to the properties dict

mode777:
I just noticed colors get parsed but get never assigned to the dict. It looks like this is also the case for some other properties as well. If you agree this is legit I can investigate and possible append more commits.


RhapsodyInGeek:
Didn't see this PR, sorry about that. Fixed this bug and a few other property assignment bugs involving the vector integer properties under f05e0aa . Thanks for the help though!

PR #6, opened 2024-04-06
Create .gitignore to comply with Godot Asset Library requirements

danila-zol:
The Asset Library application for FuncGodot has been [rejected](https://godotengine.org/asset-library/asset/edit/11253) because the repo lacks a `.gitignore`. I have added the template `.gitignore` file from the official docs. After this change the package can be resubmitted for inclusion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,15 @@
+# Godot 4+ specific ignores
+.godot/
+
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
Looks good. Sorry for the oversight!

PR #8, opened 2024-04-15
Expose texture information for solids classes so they can be used in scripting

RisingThumb:
Hi, this is just a draft PR from my fork, so I can easily rebase this change I'm using on future releases of Func_Godot.

What this does, is it exposes the textures used by a solid class, into a list of textures. This is then selectively applied to the entity properties of the solid class, so in your solid classes, you can check if a brush has a texture and if it does, do something.

The main motivation behind this change, was I needed a way to infer what textures a brush has. If it had a texture with a name like `gray_stone_bricks` and I wanted to play stone footsteps instead of default grass footstep, I can then just do a check against the staticBody which is this class, see that it has this texture present, and then play the right sound. A lot of this change, could be done just by adding a key to my solid class, and then setting it for each brush that I want to do it, but if I know I only do this when it's a particular texture, that's a lot of extra work that is error-prone that isn't necessary and also not as intuitive(in my personal case, I use it to check for the existence of a texture, then apply a PhysicsMaterial resource that I've extended with extra fields to handle audio things).

This could probably be used for further things. Only want the player to have ice physics? Check if a brush has an ice texture on it, and then do ice physics on it.

This could also probably be extended by exposing further brush information to get brush face information. Ideally I think a solid class should have some way of getting solid-specific information in the entity properties for your textures or brush information. Could also be done as an opt-in thing(similar to how the applying rotation on build is opt-out), where by default you don't expose this information unless the person making the class actually cares about these properties and wants them included.

This would also be a way that can close the gap between visual meshes and collisions. See this related Godot Proposal https://github.com/godotengine/godot-proposals/issues/3909 (interesting the proposer was using Qodot here)
There is also this article that talks about this, which showcases a lot of ways that I don't think are really that ideal to be doing for this sort of thing
http://www.footnotesforthefuture.com/words/godot-identifying-materials/

### DIFF
--- a/addons/func_godot/src/core/func_godot.gd
+++ b/addons/func_godot/src/core/func_godot.gd
@@ -49,13 +49,10 @@ func get_entity_dicts() -> Array:
 		dict["properties"] = entity.properties
 		
 		var textures_used_by_entity:Array = []
-		for i in range(entity.brushes.size()):
-			var faces = entity.brushes[i].faces
-			for j in range(faces.size()):
-				var texture_idx = faces[j].texture_idx
-				var texture_name = map_data.textures[texture_idx].name
-				if !textures_used_by_entity.has(texture_name):
-					textures_used_by_entity.append(texture_name)
+		for texture_id in entity.texture_ids:
+			var texture_to_add = map_data.textures[texture_id].name
+			if !textures_used_by_entity.has(texture_to_add):
+				textures_used_by_entity.append(texture_to_add)
 		dict["textures_used"] = textures_used_by_entity
 		
 		ent_dicts.append(dict)
--- a/addons/func_godot/src/core/func_godot.gd
+++ b/addons/func_godot/src/core/func_godot.gd
@@ -48,6 +48,16 @@ func get_entity_dicts() -> Array:
 		dict["center"] = Vector3(entity.center.y, entity.center.z, entity.center.x)
 		dict["properties"] = entity.properties
 		
+		var textures_used_by_entity:Array = []
+		for i in range(entity.brushes.size()):
+			var faces = entity.brushes[i].faces
+			for j in range(faces.size()):
+				var texture_idx = faces[j].texture_idx
+				var texture_name = map_data.textures[texture_idx].name
+				if !textures_used_by_entity.has(texture_name):
+					textures_used_by_entity.append(texture_name)
+		dict["textures_used"] = textures_used_by_entity
+		
 		ent_dicts.append(dict)
 	
 	return ent_dicts
--- a/addons/func_godot/src/core/func_godot.gd
+++ b/addons/func_godot/src/core/func_godot.gd
@@ -5,8 +5,8 @@ var map_parser:= FuncGodotMapParser.new(map_data)
 var geo_generator = preload("res://addons/func_godot/src/core/func_godot_geo_generator.gd").new(map_data)
 var surface_gatherer:= FuncGodotSurfaceGatherer.new(map_data)
 
-func load_map(filename: String, keep_tb_groups: bool) -> void:
-	map_parser.load_map(filename, keep_tb_groups)
+func load_map(filename: String, keep_tb_groups: bool, entity_definitions: Dictionary) -> void:
+	map_parser.load_map(filename, keep_tb_groups, entity_definitions)
 
 func get_texture_list() -> PackedStringArray:
 	var g_textures: PackedStringArray
@@ -49,11 +49,12 @@ func get_entity_dicts() -> Array:
 		dict["properties"] = entity.properties
 		
 		var textures_used_by_entity:Array = []
-		for texture_id in entity.texture_ids:
-			var texture_to_add = map_data.textures[texture_id].name
-			if !textures_used_by_entity.has(texture_to_add):
-				textures_used_by_entity.append(texture_to_add)
-		dict["textures_used"] = textures_used_by_entity
+		if entity.texture_ids && entity.texture_ids.size() > 0: # May be initialised as null, or empty array
+			for texture_id in entity.texture_ids:
+				var texture_to_add = map_data.textures[texture_id].name
+				if !textures_used_by_entity.has(texture_to_add):
+					textures_used_by_entity.append(texture_to_add)
+			dict["textures_used"] = textures_used_by_entity
 		
 		ent_dicts.append(dict)
 	
--- a/addons/func_godot/src/core/func_godot_map_data.gd
+++ b/addons/func_godot/src/core/func_godot_map_data.gd
@@ -98,6 +98,7 @@ class FuncGodotBrush:
 class FuncGodotEntity:
 	var properties: Dictionary
 	var brushes: Array[FuncGodotBrush]
+	var texture_ids: Array[int]
 	var center: Vector3
 	var spawn_type: FuncGodotEntitySpawnType
 	var origin_type: FuncGodotEntityOriginType
--- a/addons/func_godot/src/core/func_godot_map_parser.gd
+++ b/addons/func_godot/src/core/func_godot_map_parser.gd
@@ -13,6 +13,7 @@ var valve_uvs: bool = false
 var current_face: FuncGodotMapData.FuncGodotFace
 var current_brush: FuncGodotMapData.FuncGodotBrush
 var current_entity: FuncGodotMapData.FuncGodotEntity
+var entity_definitions: Dictionary
 
 var map_data: FuncGodotMapData
 var _keep_tb_groups: bool = false
@@ -20,10 +21,11 @@ var _keep_tb_groups: bool = false
 func _init(in_map_data: FuncGodotMapData) -> void:
 	map_data = in_map_data
 
-func load_map(map_file: String, keep_tb_groups: bool) -> bool:
+func load_map(map_file: String, keep_tb_groups: bool, entity_definitions: Dictionary) -> bool:
 	current_face = FuncGodotMapData.FuncGodotFace.new()
 	current_brush = FuncGodotMapData.FuncGodotBrush.new()
 	current_entity = FuncGodotMapData.FuncGodotEntity.new()
+	self.entity_definitions = entity_definitions;
 	
 	scope = FuncGodotMapParser.ParseScope.FILE
 	comment = false
@@ -287,6 +289,7 @@ func commit_entity() -> void:
 	new_entity.spawn_type = FuncGodotMapData.FuncGodotEntitySpawnType.ENTITY
 	new_entity.properties = current_entity.properties
 	new_entity.brushes = current_entity.brushes
+	# If empty array or null, is defined as empty array or null and we check this later on
 	new_entity.texture_ids = current_entity.texture_ids
 	map_data.entities.append(new_entity)
 	
@@ -304,7 +307,13 @@ func commit_face() -> void:
 	current_face.is_valve_uv = valve_uvs
 	
 	current_brush.faces.append(current_face)
-	current_entity.textureIds.append(current_face.texture_idx)
+	if 'classname' in current_entity.properties:
+		var classname: String = current_entity.properties['classname']
+		if classname in entity_definitions:
+			var entity_definition: FuncGodotFGDEntityClass = entity_definitions[classname] as FuncGodotFGDEntityClass
+			if entity_definition is FuncGodotFGDSolidClass:
+				if entity_definition.pass_texture_list_to_metadata:
+					current_entity.texture_ids.append(current_face.texture_idx)
 	current_face = FuncGodotMapData.FuncGodotFace.new()
 
 # Nested
--- a/addons/func_godot/src/core/func_godot_map_parser.gd
+++ b/addons/func_godot/src/core/func_godot_map_parser.gd
@@ -287,6 +287,7 @@ func commit_entity() -> void:
 	new_entity.spawn_type = FuncGodotMapData.FuncGodotEntitySpawnType.ENTITY
 	new_entity.properties = current_entity.properties
 	new_entity.brushes = current_entity.brushes
+	new_entity.texture_ids = current_entity.texture_ids
 	map_data.entities.append(new_entity)
 	
 	current_entity = FuncGodotMapData.FuncGodotEntity.new()
@@ -303,6 +304,7 @@ func commit_face() -> void:
 	current_face.is_valve_uv = valve_uvs
 	
 	current_brush.faces.append(current_face)
+	current_entity.textureIds.append(current_face.texture_idx)
 	current_face = FuncGodotMapData.FuncGodotFace.new()
 
 # Nested
--- a/addons/func_godot/src/fgd/func_godot_fgd_solid_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_solid_class.gd
@@ -52,6 +52,8 @@ enum CollisionShapeType {
 @export var collision_shape_margin: float = 0.04
 
 @export_group("Scripting")
+## Option to include the texture list in the metadata of a solid class
+@export var pass_texture_list_to_metadata: bool = false
 ## An optional script file to attach to the node generated on map build.
 @export var script_class: Script
 
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -386,6 +386,7 @@ func build_entity_nodes() -> Array:
 					node_name = "entity_" + name_prop
 				
 				if entity_definition is FuncGodotFGDSolidClass:
+					properties['textures_used'] = entity_dict['textures_used']
 					if entity_definition.spawn_type == FuncGodotFGDSolidClass.SpawnType.MERGE_WORLDSPAWN:
 						entity_nodes.append(null)
 						continue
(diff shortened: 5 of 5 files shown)


RhapsodyInGeek:
I remember seeing you discuss this on the discord. It's a neat idea, and a useful one, but I'm admittedly not sold on the implementation.

For one, there shouldn't be a need to loop through the faces of the mesh to get the texture list. There is already a texture list provided to... I believe SurfaceGatherer, so it already performs this loop (though I can't recall where exactly, it's a big plugin). I'd prefer reusing this texture list as it already gets retrieved. That way we don't unnecessarily double the amount of loops on build.

I also don't want to add that list to the entity properties that gets passed to the generated node. Instead it should be passed to the generated node's metadata so as not to conflict with a user's potential naming conventions for their own designs. There's an example of this in the code that handles TrenchBroom hierarchy.

> Could also be done as an opt-in thing(similar to how the applying rotation on build is opt-out), where by default you don't expose this information unless the person making the class actually cares about these properties and wants them included.

I would require "opt-in" to accept this proposal, since it effectively would add a Dictionary to every single solid entity otherwise and I'd rather users have control over their memory footprint as much as possible. I'd recommend adding a boolean on the **FuncGodotFGDSolidClass**, call it `pass_texture_list_to_metadata`.

RisingThumb:
I'll see if I can put some time in this week to make these changes. Out of curiosity, do you think there's value in expanding it to cover brushes as well? Could also just create an open issue just to have it mentioned, and if there's someone who has a usecase or need for it, leave it up to them to make a PR against it

I can see it being useful if you need to know the specific brush face you collided with(i.e. for brushes where there's multiple textures that do something special, like different footstep sound effects or etc), though I haven't had time to think this through or have a particular need for this in my current project

RhapsodyInGeek:
> Out of curiosity, do you think there's value in expanding it to cover brushes as well?

If you're talking about the brush data from the map, no. Not "as is", anyway. The brush face information only contains plane coordinates, texture names, and UV coordinates. The plane data itself requires a coordinate system conversion to be transformed into vertices for the generated meshes.

Trying to do a "per texture surface type" for footsteps / bulletholes / etc... is also inherently fighting against how Godot itself works, since collision data is completely detached from visual data. There are a number of far better methods to go about material types without trying to retrieve specific faces that are more performant and less error prone.

@EMBYRDEV has talked about trying to split collision objects based upon materials. That'd be something I think would be infinitely more useful for a goal like that.

RisingThumb:
> For one, there shouldn't be a need to loop through the faces of the mesh to get the texture list. There is already a texture list provided to... I believe SurfaceGatherer, so it already performs this loop (though I can't recall where exactly, it's a big plugin). I'd prefer reusing this texture list as it already gets retrieved. That way we don't unnecessarily double the amount of loops on build.

I looked into this. This is correct, SurfaceGatherer has a list of textures but that's because map_data has a list of textures, but both are across the entire map, and the functions in SurfaceGatherer take in an entity id and brush id to do their work. I can get the entity ID but I'd still need to loop over all the brushes to get textures for them.

That said, I took a look at another way of doing this. Instead of another loop, we can just keep track of the texture ids we're using when parsing the map when we commit the faces/entities, and then just go from the IDs to the names when getting the textures used. Probably should do the .has(textureId) check in there instead, but left that as is for now. Not super happy about this solution as it increases memory requirement from keeping this texture array for all entities, when it's only relevant to those that are opted in for. I'll take another look at this to see if I can only keep track of this information for entities we actually opt in for

RisingThumb:
> I'll take another look at this to see if I can only keep track of this information for entities we actually opt in for

Looked into this. The main issue here is when it's committing entities, these entities are of the type `FuncGodotEntity` defined in func_godot_map_data.gd, while the flag we set is `pass_texture_list_to_metadata` which is on the type `FuncGodotFGDSolidClass` which extends `FuncGodotFGDEntityClass`.

Is there an easy way to get the SolidClass related to an Entity? As far as I can tell, I'd have to get a list of entityDefinitions, and then get the entity's classname, match that to the definition to get the appropriate FuncGodotFGDEntity, get the solid class and then check the flag on that(similar to what's done in build_entity_nodes() in func_godot_map.gd ), though this is a lot of extra work and checks to avoid storing a list of integers only during building(though it could effect runtime building as it's a slight extra cost for each entity when runtime building). What do you think of this? @RhapsodyInGeek

RisingThumb:
> I would require "opt-in" to accept this proposal, since it effectively would add a Dictionary to every single solid entity otherwise and I'd rather users have control over their memory footprint as much as possible. I'd recommend adding a boolean on the FuncGodotFGDSolidClass, call it pass_texture_list_to_metadata.

I have made this change. This flag is checked in 2 places. Firstly in whether to attach the texture ids to the entity, and secondly in whether to record in the node's metadata. The former requires reordering the build order so we get the entity definitions first, so that I can then check this dict in the parser to see if it does have the flag set.

Personally, I'm not a big fan of this, but besides a huge refactor that changes how the map parsing is done, to use the actual entity definitions and not this sort of "in between" entity that's used for generation, I don't really see how else we can do this. The consequences of this are 4-5 extra checks against a dict for entity committed.

I would be curious to hear if there's any regressions or issues that occur as a result of getting the entity definitions as an earlier build step. Moving the following build step up with it: `register_build_step('set_func_godot_entity_definitions', [])` results in mangled world geometry, however moving this isn't necessary for what I'm doing here. I suspect this is because that setter does a lot of extra non-trivial logic

RhapsodyInGeek:
Sorry I haven't been as responsive to this. Been pretty busy. I don't have answers yet, still parsing what you've done and need to look over how it all works again. Admittedly, also not really as much of a priority to me as #15 but it is somewhat adjacent to it in the sense that I have to research how GeoGenerator and SurfaceGatherer work and both #15 and this are tied to those two pieces of core.

I'll see about looking more into it at some point tomorrow. I can see the potential usefulness of a texture list, I'm just not sold on this implementation until I'm more confident in my understanding of GeoGen and SurfaceGatherer.

DeerTears:
This looks pretty alright to me! I still need to test this, but it looks very functional for a new hook. It's enough to get people like you started with your own material system.

I am a bit worried that this work - while useful now - will probably get superceded when more work is done on separating the map into per-material bodies, what RhapsodyInGeek and EMBYRDEV were talking about. I only say that because your system (as far as I can see) assumes you'll be splitting your map into brush entities per material, and then also writing a script to modify the brush entities after the map is built. That's not bad for some mappers, but this system isn't flexible to fit with all mappers' styles. Hopefully when we do roll out something more robust, it'll satisfy all mappers who want to have per-face footsteps and decals, not just per-brush footsteps and decals. (Unless I'm misunderstanding something big about this PR!)

I still want to work toward getting this merged, knowing that it would help some users now. We are otherwise volunteers and a feature like this would still be sorely wanted in the meantime while we're figuring it out. And I don't see the damage that could be done by introducing it to see how people use the system, and if we can take any cues from that to improve upon it in upcoming versions.

### I have some questions about this PR!
1. Does calling `load_map()` correctly update the FGD when `verify_and_build()` is used at runtime? Or should this new argument be referring to what's the in FuncGodotMapSettings instead of being an argument? If you don't know, I'll just test for this.
2. Can there be a default setting for that new argument in `load_map()`? Ideally, using what's already loaded in the map if the argument is left blank. Then it's backwards compatible if people are using `load_map()` for runtime building. An example would be `func load_map(file, use_trenchbroom_hierarchy, entity_definitions: Dictionary = self.func_godot_map_settings.entity_fgd.entity_definitions)`.
3. How were you planning to check all of these texture names for every affected brush? Can we go a step further in making this hook reachable, and add all opt-in brush entities to a scene-wide group? Like `add_to_group("fgd_has_texture_list")`. That way it'd be easier for any script to apply these material properties to each affected brush entity, without modifying each brush entity's script.

I'm excited to test this out. :3 Let me know your thoughts.

RisingThumb:
> This system isn't bad for mappers who can adapt their mapping style to meet this system's needs, like ensuring all faces of a brush have the same texture to be interpreted as the same material. But in general I don't think this system is flexible to fit with all mappers' styles. So hopefully when we do roll out something more robust, it'll satisfy all mappers who want to have per-face footsteps and decals, not just per-brush footsteps and decals. (Unless I'm misunderstanding something big about this PR!)

This system doesn't enforce all brushes having the same texture being interpreted as the same material. It just provides a list of textures a brush entity features. What a person does with the texture list is up to them. As an example of what I'm doing with it... physics material-based footsteps:
```
	if has_meta("textures_used"):
		textures_used = get_meta("textures_used");
	if !self.mapper.DefaultPhysicsMaterial:
		push_error("No default physics material set for the material mapper in func_geo...")
		return
	if !self.mapper.ExpandedPhysicsMaterialDictionary:
		push_error("No mapper for physics material set for the material mapper in func_geo...")
		return
	self.physics_material_override = self.mapper.DefaultPhysicsMaterial
	for texture:String in textures_used:
		for key:String in self.mapper.ExpandedPhysicsMaterialDictionary.keys():
			if texture.contains(key):
				self.physics_material_override = self.mapper.ExpandedPhysicsMaterialDictionary[key]
```
As static bodies have a limit of 1 physics material, this bakes an assumption into my workflow, but this isn't really relevant to func_godot. You can drive a bunch of properties automatically by the textures. If it has an ice texture, you can make it do weird ice physics. If it has a lava texture, you can make it hurt e.t.c. Personally, I think it helps drive automatic property assignment by texture instead of setting them all in-editor(but to each their own how they want to use this texture list).

> Does calling load_map() correctly update the FGD when verify_and_build() is used at runtime? Or should this new argument be referring to what's the in FuncGodotMapSettings instead of being an argument? If you don't know, I'll just test for this.

In the testing I've done, yes. `verify_and_build()` registers all the build steps and then runs through them all, so load_map is called after getting all the entity dictionaries. load_map is just one of the build steps. I can also confirm this works with runtime map building.

> Or should this new argument be referring to what's the in FuncGodotMapSettings instead of being an argument?

I don't follow what's meant here. `fetch_entity_definitions()` already fetches them from the FuncGodotMapSettings's FGD if that's what you're asking.

> Can there be a default setting for that new argument in load_map()? Right now if someone was using load_map() at runtime they'd receive an error for missing a new argument. An example fix would be func load_map(file, use_trenchbroom_hierarchy, entity_definitions: Dictionary = self.func_godot_map_settings.entity_fgd.entity_definitions).

Sure, I could default it to an empty dict, or default it to `self.func_godot_map_settings.entity_fgd.entity_definitions()` but keep in mind, that's a function call which does a bunch of logic that'd be getting repeated here(so not a super big fan of that choice). Looking at it, I think moving our entity definitions getting to the top of the build order is correct because the Map Setting FGD doesn't change during build so the FGD entity definitions don't change during build.

> How were you planning to check all of these texture names for every affected brush? Can we go a step further in making this hook reachable, and add all opt-in brush entities to a scene-wide group? Like add_to_group("fgd_has_texture_list"). That way it'd be easier for any script to apply these material properties to each affected brush entity, without modifying each brush entity's script.

I'm not sure I follow. By default it's turned off, because I don't think this functionality is useful to everyone(and adds extra memory costs to have it enabled). If someone wants it added to a group, they could do so in their brush entity's tool script, to check for the presence of a texture list and add it to a group. Apologies if it's a non-answer, I don't think I follow.

> I am a bit worried that this work - while useful now - will probably get superceded when more work is done on separating the map into per-material bodies, what RhapsodyInGeek and EMBYRDEV were talking about. I only say that because your system (as far as I can see) assumes you'll be splitting your map into brush entities per material, and then also writing a script to modify the brush entities after the map is built.

Can you link the discussion on this? I think this will work alongside it as it just provides the texture list for a brush entity and doesn't do anything related to materials or brushes or faces. If it gets overridden or tossed out in a refactor that does this better, I'm happy for that too(though I imagine such a change may be compatibility-breaking). That said, this is all speculation without any discussion I've read up on yet.

> and then also writing a script to modify the brush entities after the map is built.

I think this is the point. On their own the texture list isn't useful unless you're gonna do something with it

RhapsodyInGeek:
> Does calling load_map() correctly update the FGD when verify_and_build() is used at runtime?

`load_map()` gets called during every build. `verify_and_build()` doesn't distinguish between Editor Time and Runtime building. `load_map()` is the function that actually begins the map parsing process by loading the map file data into the MapParser core.

> I think moving our entity definitions getting to the top of the build order is correct because the Map Setting FGD doesn't change during build so the FGD entity definitions don't change during build.

This is a good suggestion.

> This system isn't bad for mappers who can adapt their mapping style to meet this system's needs, like ensuring all faces of a brush have the same texture to be interpreted as the same material.

I know RT had already touched on this, but one way I'd probably do it is tally the number of faces that each texture in the texture list covers, then the one with the most coverage gets the associated PhysicsMaterial applied to it. This would be done at build time of course. But yeah, not something func_godot needs to worry about, that's more of a general design thing.

> Can you link the discussion on this?

There isn't a proper discussion on it yet, but it was one of the impetuses that instigated Embyr to go through with a func_godot rewrite. No headway has been made on this as of yet.

RhapsodyInGeek:
I'm currently working on a Core cleanup / small refactor, and I think I see an opportunity to not only get the used textures list cleanly, but also get a count on how many faces are using that texture. Might be helpful to get the texture face count to help automate certain physics material application methods (ex: material applied goes to most used texture).

I can send the solid class info over through `set_func_godot_entity_definitions()` like how spawn type and origin type get passed through to GeoGen and SurfaceGatherer.

RisingThumb:
> I'm currently working on a Core cleanup / small refactor, and I think I see an opportunity to not only get the used textures list cleanly, but also get a count on how many faces are using that texture. Might be helpful to get the texture face count to help automate certain physics material application methods (ex: material applied goes to most used texture).
> 
> I can send the solid class info over through `set_func_godot_entity_definitions()` like how spawn type and origin type get passed through to GeoGen and SurfaceGatherer.

@RhapsodyInGeek I've been thinking a little bit about this. I'll change what gets stored in the metadata from a list to a dictionary, where the key is the texture name and the value is the number of occurrences of that texture(which also means if you can easily just get the array again by doing .keys() without duplicates). I'm thinking half of the approach here is still the correct way to do things: The part about it being applied as a flag on solid classes(to ease memory cost for built maps), and the part about attaching it as metadata.

When I take another look at this, probably later this week, which part of the codebase should I be looking at hooking this into? Into the `set_func_godot_entity_definitions()` method?

RhapsodyInGeek:
> The part about it being applied as a flag on solid classes(to ease memory cost for built maps), and the part about attaching it as metadata.

That's definitely the approach I was thinking of. When I meant "not sold on the implementation" what I meant was where it was happening. I agree the filter of a Solid Class flag and the result of a textures Dictionary sent to the node's meta data are the right way to go about it.

> When I take another look at this, probably later this week, which part of the codebase should I be looking at hooking this into? Into the set_func_godot_entity_definitions() method?

The FuncGodot Core (`src/core/func_godot.gd`) has this method:
```gdscript
func set_entity_definitions(entity_defs: Dictionary) -> void:
	for i in range(entity_defs.size()):
		var classname: String = entity_defs.keys()[i]
		var spawn_type: int = entity_defs.values()[i].get("spawn_type", FuncGodotMapData.FuncGodotEntitySpawnType.ENTITY)
		var origin_type: int = entity_defs.values()[i].get("origin_type", FuncGodotMapData.FuncGodotEntityOriginType.IGNORE)
		map_data.set_entity_types_by_classname(classname, spawn_type, origin_type)
```

Then the map_data object calls this:
```gdscript
func set_entity_types_by_classname(classname: String, spawn_type: int, origin_type: int) -> void:
	for entity in entities:
		if entity.properties.has("classname") and entity.properties["classname"] == classname:
			entity.spawn_type = spawn_type as FuncGodotMapData.FuncGodotEntitySpawnType
			if entity.spawn_type == FuncGodotMapData.FuncGodotEntitySpawnType.ENTITY:
				entity.origin_type = origin_type as FuncGodotMapData.FuncGodotEntityOriginType
			else:
				entity.origin_type = FuncGodotMapData.FuncGodotEntityOriginType.IGNORE
```
You might recognize these properties if you've looked around in GeoGenerator or SurfaceGatherer. They basically add the selected SolidClass definition properties to the FuncGodotEntity:
```gdscript
class FuncGodotEntity:
	var properties: Dictionary
	var brushes: Array[FuncGodotBrush]
	var center: Vector3
	var spawn_type: FuncGodotEntitySpawnType
	var origin_type: FuncGodotEntityOriginType
```
Thinking more on this though, it might be worth just passing a Dictionary through instead of adding more arguments to `FuncGodotMapData.set_entity_types_by_classname()`. We'd still have to modify the contents of each method and the FuncGodotEntity, so I guess it doesn't matter too much.

In any case, we'd pass the SolidClass flag through those functions and add a bool to `FuncGodotMapData.FuncGodotEntity` to determine whether we collect a texture list or not, and then we can check it when it gets passed through either GeoGenerator or SurfaceGatherer. Probably best during SurfaceGatherer where it sets up the surfaces for the final resulting mesh.

RisingThumb:
Following from this PR and comment https://github.com/func-godot/func_godot_plugin/pull/23#issuecomment-2241768601

> You'll already be able to access the materials list from the MeshInstance3D directly. While I'm not against this, I'm also not convinced of the usefulness when the data is already available.

I don't think this PR has much use anymore as the texture data is already available from this material(and if the material's albedo isn't the same, the material's resource name will be). Closing this PR as a result, feel free to reopen it if you think there's any benefit in anything in it

PR #11, opened 2024-04-17
Created C# Interface for FGD

ionthedev:
This interface makes it a little less cumbersome when working with FGD entities in C#.

Requires Mono build of Godot and interface needs to be added to the script. Simply write "public override _func" and let intellisense do the rest.

There was a bit of confusion that came from the documentation in relation to C# support and the Entity Class Definitions, hopefully this will mediate that a bit.


RhapsodyInGeek:
One of the core design pillars of func_godot is to keep the plugin pure GDScript. While a neat implementation, this unfortunately goes against that core design and so will not be merged.

RisingThumb:
> One of the core design pillars of func_godot is to keep the plugin pure GDScript. While a neat implementation, this unfortunately goes against that core design and so will not be merged.

Maybe I'm blind, but I haven't seen anywhere that this was a design pillar. I'll push back a little bit because it doesn't seem like much effort to add support for C#(maybe @ionthedev can add more detail on this if any extra steps are needed), and could probably be optional thing to use(if it demands the inclusion of mono, then that's obviously a negative). That said, I believe there is some benefit here just because of how many C# developers who come from Unity there are.

At the very least, I believe we should document that it's possible to use with C#(or other languages like C++ iirc from discord discussion?)

ionthedev:
> > One of the core design pillars of func_godot is to keep the plugin pure GDScript. While a neat implementation, this unfortunately goes against that core design and so will not be merged.
> 
> Maybe I'm blind, but I haven't seen anywhere that this was a design pillar. I'll push back a little bit because it doesn't seem like much effort to add support for C#(maybe @ionthedev can add more detail on this if any extra steps are needed), and could probably be optional thing to use(if it demands the inclusion of mono, then that's obviously a negative). That said, I believe there is some benefit here just because of how many C# developers who come from Unity there are.
> 
> At the very least, I believe we should document that it's possible to use with C#(or other languages like C++ iirc from discord discussion?)

I understand the core pillar philosophy, I know it can be hard to maintain things that span different languages.

The reason I wanted to develop this is because not everyone is going to know how to write a compatible function in C# to work with the system properly; Hence why I made the Interface. 

Personally though, I think purposely not including C# support will push away some users from using the addon as a whole.

RhapsodyInGeek:
> Personally though, I think purposely not including C# support will push away some users from using the addon as a whole.

This hasn't really proven true, since multiple users have been using FuncGodot with C# without issue, as had users with Qodot after Embyr converted it from a mix of C# and GDScript to pure GDScript. There's no real need for C# support in FuncGodot since FuncGodot doesn't stop you from using C#. The Godot-Mono requirement was also a sore point for most of our users, which was one of the main reasons we switched to pure GDScript.

> The reason I wanted to develop this is because not everyone is going to know how to write a compatible function in C# to work with the system properly; Hence why I made the Interface.

This is honestly applying a band aid to an education issue that isn't as large as you might think.

> If FuncGodot was written as a gdextension, It would mediate the issue for it being comparable with either language.

It was actually discussed whether or not to write FuncGodot in GDExtension (C++ GDX is something I'm a huge fan of) but the team ultimately agreed the plugin would be pure GDScript, mostly due to its portability.

PR #13, opened 2024-05-04
Adds Automatic Building Functionality to Map Nodes

Pspritechologist:
Requiring a manual press of a 'build' button after each map change to see the effects within Godot becomes tedious quickly, not to mention requiring that you manually hunt down each map Node one at a time to do so.

This PR adds an `auto_reload` property to map Nodes allowing them to react to the reimporting of a map file, making working with them very similar to working with .blend files directly.

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -40,9 +40,10 @@ var _map_file_internal: String = ""
 @export var block_until_complete: bool = false
 ## How many nodes to set the owner of, or add children of, at once. Higher values may lead to quicker build times, but a less responsive editor.
 @export var set_owner_batch_size: int = 1000
-## If true, automatically builds this map when the relevant map file is reimported.
+## If true, automatically triggers a build on this Node when the relevant local map file is imported.
+## An import can occur when the Resource is first added, when the resource is updated externally, or when an import is manually triggered.
 ## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
-@export var auto_build: bool = true
+@export var auto_build_on_local_file_import: bool = false
 
 # Build context variables
 var func_godot: FuncGodot = null
@@ -72,7 +73,7 @@ var entity_collision_shapes: Array = []
 func _ready() -> void:
 	if Engine.is_editor_hint():
 		EditorInterface.get_resource_filesystem().resources_reimported.connect(_on_reimport)
-		if auto_build:
+		if auto_build_on_local_file_import:
 			verify_and_build()
 	
 	if not DEBUG:
@@ -91,7 +92,7 @@ func verify_and_build() -> void:
 		emit_signal("build_failed")
 
 func _on_reimport(resources: PackedStringArray) -> void:
-	if !auto_build:
+	if !auto_build_on_local_file_import:
 		return
 	
 	for resource in resources:
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -43,7 +43,7 @@ var _map_file_internal: String = ""
 ## If true, automatically triggers a build on this Node when the relevant local map file is imported.
 ## An import can occur when the Resource is first added, when the resource is updated externally, or when an import is manually triggered.
 ## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
-@export var auto_build_on_local_file_import: bool = false
+@export var auto_build_on_local_file_import: bool = true
 
 # Build context variables
 var func_godot: FuncGodot = null
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -40,10 +40,8 @@ var _map_file_internal: String = ""
 @export var block_until_complete: bool = false
 ## How many nodes to set the owner of, or add children of, at once. Higher values may lead to quicker build times, but a less responsive editor.
 @export var set_owner_batch_size: int = 1000
-## If true, automatically triggers a build on this Node when the relevant local map file is imported.
-## An import can occur when the Resource is first added, when the resource is updated externally, or when an import is manually triggered.
-## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
-@export var auto_build_on_local_file_import: bool = true
+## If true, automatically rebuilds the map when the local map file is reimported after being saved externally.
+@export var auto_build_on_local_map_update: bool = true
 
 # Build context variables
 var func_godot: FuncGodot = null
@@ -73,7 +71,7 @@ var entity_collision_shapes: Array = []
 func _ready() -> void:
 	if Engine.is_editor_hint():
 		EditorInterface.get_resource_filesystem().resources_reimported.connect(_on_reimport)
-		if auto_build_on_local_file_import:
+		if auto_build_on_local_map_update:
 			verify_and_build()
 	
 	if not DEBUG:
@@ -92,7 +90,7 @@ func verify_and_build() -> void:
 		emit_signal("build_failed")
 
 func _on_reimport(resources: PackedStringArray) -> void:
-	if !auto_build_on_local_file_import:
+	if !auto_build_on_local_map_update:
 		return
 	
 	for resource in resources:
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -40,6 +40,16 @@ var _map_file_internal: String = ""
 @export var block_until_complete: bool = false
 ## How many nodes to set the owner of, or add children of, at once. Higher values may lead to quicker build times, but a less responsive editor.
 @export var set_owner_batch_size: int = 1000
+## If true, automatically builds this map when the relevant map file is reimported.
+## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
+@export var auto_build: bool = true:
+	set(value):
+		auto_build = value
+		if auto_build && !EditorInterface.get_resource_filesystem().resources_reimported.is_connected(_on_reimport):
+			EditorInterface.get_resource_filesystem().resources_reimported.connect(_on_reimport)
+		
+		if !auto_build && EditorInterface.get_resource_filesystem().resources_reimported.is_connected(_on_reimport):
+			EditorInterface.get_resource_filesystem().resources_reimported.disconnect(_on_reimport)
 
 # Build context variables
 var func_godot: FuncGodot = null
@@ -67,6 +77,9 @@ var entity_collision_shapes: Array = []
 
 # Overrides
 func _ready() -> void:
+	if Engine.is_editor_hint() && auto_build:
+		verify_and_build()
+	
 	if not DEBUG:
 		return
 	
@@ -81,6 +94,11 @@ func verify_and_build() -> void:
 		build_map()
 	else:
 		emit_signal("build_failed")
+
+func _on_reimport(resources: PackedStringArray) -> void:
+	for resource in resources:
+		if resource == local_map_file:
+			verify_and_build()
 
 ## Build the map.
 func manual_build() -> void:
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -42,16 +42,7 @@ var _map_file_internal: String = ""
 @export var set_owner_batch_size: int = 1000
 ## If true, automatically builds this map when the relevant map file is reimported.
 ## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
-@export var auto_build: bool = true:
-	set(value):
-		if !Engine.is_editor_hint():
-			return
-		auto_build = value
-		if auto_build && !EditorInterface.get_resource_filesystem().resources_reimported.is_connected(_on_reimport):
-			EditorInterface.get_resource_filesystem().resources_reimported.connect(_on_reimport)
-		
-		if !auto_build && EditorInterface.get_resource_filesystem().resources_reimported.is_connected(_on_reimport):
-			EditorInterface.get_resource_filesystem().resources_reimported.disconnect(_on_reimport)
+@export var auto_build: bool = true
 
 # Build context variables
 var func_godot: FuncGodot = null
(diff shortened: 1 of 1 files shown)


Pspritechologist:
Pretty new to the repo, but I think this tool is fantastic! When I found that I'd need to hit 'build' after every map change, I already knew I'd need to implement something to change it. When I found that I could implement it so easily though, I figured I may as well see if anyone else found it useful as well :P

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -82,6 +95,11 @@ func verify_and_build() -> void:
 	else:
 		emit_signal("build_failed")
 
+func _on_reimport(resources: PackedStringArray) -> void:
+	for resource in resources:
+		if resource == local_map_file:
```
RisingThumb, line 102: This should also be checking global_map_file

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -82,6 +95,11 @@ func verify_and_build() -> void:
 	else:
 		emit_signal("build_failed")
 
+func _on_reimport(resources: PackedStringArray) -> void:
+	for resource in resources:
```
RisingThumb, line 101: An alternative is to connect it in our ready block, and do an autobuild guard check here.
```
if !auto_build:
    return;
```
That said, each reimport would still hit this function.

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -67,6 +77,9 @@ var entity_collision_shapes: Array = []
 
 # Overrides
 func _ready() -> void:
+	if Engine.is_editor_hint() && auto_build:
+		verify_and_build()
```
RisingThumb, line 83: This will build every single time this scene is opened in the editor, which means you'll be hit with a lag spike every time you open a map as it builds. Can we do something else here?

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -82,6 +95,11 @@ func verify_and_build() -> void:
 	else:
 		emit_signal("build_failed")
 
+func _on_reimport(resources: PackedStringArray) -> void:
+	for resource in resources:
+		if resource == local_map_file:
+			verify_and_build()
```
RisingThumb, line 103: Should/Can we have an additional option to autobuild UV2s alongside this.
This UV2 unwrapping might not be necessary though because I think it will emit the finished building signal we can listen to

RisingThumb:
This functionality looks really useful, especially for people who have a workflow going in and out of the editor 👍

addons/func_godot/src/map/func_godot_map.gd
Pspritechologist: Not possible, Godot won't notify about a file change that *isn't int its directory*. If the map *is* in the directory, I don't see why the global field is being used at all.
If I'm misunderstanding the use case, please let me know.

addons/func_godot/src/map/func_godot_map.gd
Pspritechologist: If you'd like, but that's part of the point for me. If you've got a particularly large map that you don't want to generate every time you open it, you can simply turn the bool off for it. 

Without getting far more complicated with it, there's no way to check if a Resource *has been* modified since the last time a Scene was open, and so simply rebuilding each time it's opened seemed minor to me in favour of it being up to date. I was impressed with how performant the plugin has been since I started using it, but I have only used extremely simple maps. 

On a side note, this PR *doesn't* deal with the requirement to open each Scene containing a map file you've modified each time you alter them. Batch editing maps seems like it could grow to be a huge pain, but I don't know a solution personally. You'd need to find map Nodes and load the Scene up in the background before re-saving the file each time the Resource is edited, which sounds a little sketchy. Funnily enough, it'd also solve the problem listed here :P

addons/func_godot/src/map/func_godot_map.gd
Pspritechologist: I'll be honest, haven't touched that part of the plugin at all, so I have no idea :P
I could go read up on that part of the docs if you consider it important.

addons/func_godot/src/map/func_godot_map.gd
Pspritechologist: Ah, that's a good point. It would write cleaner that way, rather than fiddling with signals all over.

addons/func_godot/src/map/func_godot_map.gd
RhapsodyInGeek: UV2 unwrapping would have to yield to build completion, and even then the only GI that requires UV2 unwrapping is LightmapGI. You'd need an additional `auto_unwrap_uv2`, which could be handy anyway. It should _not_ be reliant on `auto_build` though, and instead be a completely separate consideration in the build process.

addons/func_godot/src/map/func_godot_map.gd
RisingThumb: > Not possible, Godot won't notify about a file change that isn't inside the project directory. If the map is in the directory, I don't see why the global field is being used at all.

I'm more concerned about user error here. global_map_file is supposed to override local_map_file, that logic should be here too. Additionally, it catches cases where people are using global_map_file field for local maps. View the class reference on FuncGodotMap for this overriding, specifically on the properties

addons/func_godot/src/map/func_godot_map.gd
RisingThumb: It's not currently that useful since lightmap baking isn't exposed for script use currently by Godot- which is the point of the UV2 unwrapping so you have UV2s to use for the lightmap baking.

However, when it's exposed, this will be useful. This is discussed in runtime map baking, and if it's easy to add this as an option, I think it should be done. Otherwise, don't worry too much about it

addons/func_godot/src/map/func_godot_map.gd
RisingThumb: Perhaps it should be moved from a func_godot_map field, to a func_godot_map_settings field, this way it applies to all the maps using these settings

RhapsodyInGeek:
It's a neat feature, probably useful for some folks.

- The variable should be renamed to `auto_rebuild_on_local_map_update`. It's a lot more verbose, but the verbosity also tells us exactly what it's supposed to do, and its limitation.
- The default value should be set to `false`.
- Description probably better as `## If true, automatically rebuilds the map when the local map file is reimported after being saved externally.`

Saw a question about the purpose of global map files. There are actually a lot of use cases for them, particularly for runtime building / modding. I do not see the auto-rebuild feature needing to support that though, so the limitation to local files is perfectly acceptable as long as we tell users that's a limitation.

Any issues if you don't have a map file set?

Pspritechologist:
> Saw a question about the purpose of global map files. There are actually a lot of use cases for them, particularly for runtime building / modding. I do not see the auto-rebuild feature needing to support that though, so the limitation to local files is perfectly acceptable as long as we tell users that's a limitation.
> 
> Any issues if you don't have a map file set?

Ah, runtime loading makes a ton of sense for global path!

An imported Resource will never equal an empty string, so should be no issue at all.
Even if it could trigger an auto build, it should be about as problematic as clicking the build button on a Node with no map file set.

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -40,6 +40,9 @@ var _map_file_internal: String = ""
 @export var block_until_complete: bool = false
 ## How many nodes to set the owner of, or add children of, at once. Higher values may lead to quicker build times, but a less responsive editor.
 @export var set_owner_batch_size: int = 1000
+## If true, automatically builds this map when the relevant map file is reimported.
+## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
+@export var auto_build: bool = true
```
Pspritechologist, line 45:
```suggestion
@export var auto_build_on_local_file_import: bool = false
```

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -40,6 +40,9 @@ var _map_file_internal: String = ""
 @export var block_until_complete: bool = false
 ## How many nodes to set the owner of, or add children of, at once. Higher values may lead to quicker build times, but a less responsive editor.
 @export var set_owner_batch_size: int = 1000
+## If true, automatically builds this map when the relevant map file is reimported.
+## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
```
Pspritechologist, line 44:
```suggestion
## If true, automatically triggers a build on this Node when the relevant local map file is imported.
## An import can occur when the Resource is first added, when the resource is updated externally, or when an import is manually triggered.
## This means saving over the map file and tabbing into Godot will cause the map to update automatically.
```

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -67,6 +70,11 @@ var entity_collision_shapes: Array = []
 
 # Overrides
 func _ready() -> void:
+	if Engine.is_editor_hint():
+		EditorInterface.get_resource_filesystem().resources_reimported.connect(_on_reimport)
+		if auto_build:
```
Pspritechologist, line 75:
```suggestion
		if auto_build_on_local_file_import:
```

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -82,6 +90,14 @@ func verify_and_build() -> void:
 	else:
 		emit_signal("build_failed")
 
+func _on_reimport(resources: PackedStringArray) -> void:
+	if !auto_build:
```
Pspritechologist, line 94:
```suggestion
	if !auto_build_on_local_file_import:
```

Pspritechologist (commented):
Web edits so I don't need to push :sunglasses:

Pspritechologist:
> It's a neat feature, probably useful for some folks.
> 
> * The variable should be renamed to `auto_rebuild_on_local_map_update`. It's a lot more verbose, but the verbosity also tells us exactly what it's supposed to do, and its limitation.
> * The default value should be set to `false`.
> * Description probably better as `## If true, automatically rebuilds the map when the local map file is reimported after being saved externally.`

* I've used the word 'import' frequently rather than 'update' to more directly clarify what triggers the process.
* Honestly speaking, it was meant to 'false' when I first made the pr :P. I of course prefer the functionality to be default which is why I had it otherwise locally, but I'll just set that when I use it myself. :+1: 
* Does this description seem acceptable to you? It's quite a bit more verbose but I enjoy explaining the details of what 
something is, and why the user might care about it, since it displays as the tooltip. I'm happy to shorten it to what you have if you'd prefer.

addons/func_godot/src/map/func_godot_map.gd
Pspritechologist: I considered that, the trouble is that it's far more per-map considerate than any other setting in that Resource.

RhapsodyInGeek:
> I've used the word 'import' frequently rather than 'update' to more directly clarify what triggers the process.

Should still be 'update', as that's the practical application, what's going on in the layman sense: they updated their map, so the FuncGodotMap is automatically rebuilding it. Also, should be specified as "local map file" (or just "local map") because that's much more clear than just calling it "local file".

> Does this description seem acceptable to you? It's quite a bit more verbose but I enjoy explaining the details of what
something is, and why the user might care about it, since it displays as the tooltip. I'm happy to shorten it to what you have if you'd prefer.

Too much, most of it's superfluous to be honest. The description I gave you tells everything the user needs to know about what it does in a succinct way. The member variable descriptions shouldn't enter the realm of tutorials. Descriptive, but brief.

addons/func_godot/src/map/func_godot_map.gd
RhapsodyInGeek: `auto_unwrap_uv2` is still useful in the sense that if you build your map and plan on using LightmapGI you can have it auto unwrap on build instead of remembering to press the button every time. You don't need to immediately bake your lightmap upon unwrapping the UV2s. I imagine it'd be useful as a sister feature to the auto rebuild, which is also not really useful at runtime either. 

Be careful that you don't tunnel vision into your runtime building use cases when judging the utility of features: you're actually in the minority of our users who use func_godot that way.

addons/func_godot/src/map/func_godot_map.gd
RisingThumb: Perhaps... perhaps we can have one in the map settings that if set, overrides it for all maps using it as a setting? Ultimately, I'm not too fussed about it

RhapsodyInGeek: I had thought about suggesting it be in map settings, but honestly I feel it fits more on a per-map basis rather than a gamewide basis. I'd have to agree with Psprite's judgment in this specific case.

RisingThumb: I'm not 100% convinced. Currently a lot of the settings that you could argue are per-map, are in these settings. Stuff like unshaded, save generated materials, use trenchbroom groups hierarchy etc. It also makes an opinion that map settings are gamewide, which I also don't think is always the case(unshaded is a good example of this!).
- You want to use dev settings for development("mat_fullbright 1" for unshaded) and production settings for prod
- Reloading only props that are being worked on(dev/prod idea, but with prod you don't want runtime map loading in your finished exports. Not everyone wants to expose dev tools in their finished games)
- Toggle on for the map setting, open all scenes using these map settings to regenerate them, and then toggle it off in the settings(as opposed to turning it on and off in each FuncGodotMap scene if they aren't using a premade FuncGodotMap scene). If you have to turn it on manually for all maps, that will get tedious fast

While I agree it's more per-map specific, so are others like unshaded and use trenchbroom group hierarchy

RhapsodyInGeek: Fair assessment. The original intent of a map settings resource was to share setups across maps without having to re-do them every time. A side benefit of this was using the same map with different settings could yield different results, and a way to apply it to the auto_rebuild / auto_uv_unwrap features would be to have a "production map settings" and a "release map settings", or some other combination that you can hot swap in and out.

While you could have an additional "override" variable that dictates whether an auto rebuild occurs or not, I'd lean on the side of "not". While we do have some overrides scattered about here and there, we got rid of most of the ones that were found in Qodot as they were bandaids to avoid breaking projects.

I'd have to defer to @RisingThumb in this case, the auto rebuild setting does belong in the map settings resource given what it's used for already. Shouldn't be all that difficult to change over though.

RisingThumb:
Just to add, this comment is still outstanding. All the other changes look good to me, thanks for addressing these! 👍 
https://github.com/func-godot/func_godot_plugin/pull/13#discussion_r1590353609

RhapsodyInGeek:
This PR is no longer compatible with the current version of FuncGodot due to extensive changes to FuncGodotMap. That said, an alternative option based upon this PR was devised by @RisingThumb here: https://github.com/RisingThumb/func_godot_scene_import

Closing this PR as not currently planned and no longer needed.

PR #16, opened 2024-05-13
Integrates FuncGodot with Godot's Usual Settings

Pspritechologist:
This pr switches FuncGodot's local and project-wide settings to use Godot's Editor and Project Settings respectively.

Accessing local configuration options via a 'Resource' within the Plugin's directory that only exists once and is never actually modified while the real information is saved to a JSON file within the game's app data directory is- although quite an interesting solution- not really ideal. Anything used to configure how the plugin operates with the individual game or project can be exposed to the user in ProjectSettings, and anything used to configure the interaction of the plugin with the user's local machine can be done with EditorSettings, both of which can be set up by the plugin when enabled. 
This eliminates the need for the user to re-specify the location of their editor for each project, the duplicated saving of the same information, and means that the user doesn't need to access an arbitrary file hidden somewhere in the plugin directory.

This pr isn't fully finished as ideally it would *also* move per-editor configuration such as the TrenchBroom file into EditorSettings, but I'm making it as a draft now to open discussion to these changes and what others think of them. My piece would be that placing configuration files of any kind, be it default map Resources, local config, etc., into the Plugins directory is sub-optimal and should be avoided, while utilizing the configuration system Godot *already* has in place makes the addon more intuitive, easier to operate, more compatible with external tools and knowledge, and overall feel more 'first party' and cleanly integrated with the Editor.

### DIFF
--- a/addons/func_godot/src/func_godot_plugin.gd
+++ b/addons/func_godot/src/func_godot_plugin.gd
@@ -76,11 +76,11 @@ func _exit_tree() -> void:
 		func_godot_map_progress_bar = null
 	
 	# Cleanup Editor Settings
-	FuncGodotLocalConfig.remove_editor_settings()
+	#INFO: I realized that doing this defeats the purpose of using EditorSettings
+	# FuncGodotLocalConfig.remove_editor_settings()
 	
 	# Cleanup Project Settings
-	#INFO: I realized that doing this defeats the purpose of using EditorSettings
-	# FuncGodotProjectConfig.remove_project_settings()
+	FuncGodotProjectConfig.remove_project_settings()
 
 ## Create the toolbar controls for [FuncGodotMap] instances in the editor
 func create_func_godot_map_control() -> Control:
--- a/addons/func_godot/src/func_godot_plugin.gd
+++ b/addons/func_godot/src/func_godot_plugin.gd
@@ -49,6 +49,7 @@ func _enter_tree() -> void:
 	
 	# Setup Editor Settings
 	FuncGodotLocalConfig.setup_editor_settings()
+	FuncGodotLocalConfig.cleanup_legacy()
 
 func _exit_tree() -> void:
 	remove_custom_type("FuncGodotMap")
--- a/addons/func_godot/src/func_godot_plugin.gd
+++ b/addons/func_godot/src/func_godot_plugin.gd
@@ -50,6 +50,9 @@ func _enter_tree() -> void:
 	# Setup Editor Settings
 	FuncGodotLocalConfig.setup_editor_settings()
 	FuncGodotLocalConfig.cleanup_legacy()
+	
+	# Setup Project Settings
+	FuncGodotProjectConfig.setup_project_settings()
 
 func _exit_tree() -> void:
 	remove_custom_type("FuncGodotMap")
@@ -72,7 +75,11 @@ func _exit_tree() -> void:
 		func_godot_map_progress_bar.queue_free()
 		func_godot_map_progress_bar = null
 	
+	# Cleanup Editor Settings
 	FuncGodotLocalConfig.remove_editor_settings()
+	
+	# Cleanup Project Settings
+	FuncGodotProjectConfig.remove_project_settings()
 
 ## Create the toolbar controls for [FuncGodotMap] instances in the editor
 func create_func_godot_map_control() -> Control:
--- a/addons/func_godot/src/func_godot_plugin.gd
+++ b/addons/func_godot/src/func_godot_plugin.gd
@@ -79,7 +79,8 @@ func _exit_tree() -> void:
 	FuncGodotLocalConfig.remove_editor_settings()
 	
 	# Cleanup Project Settings
-	FuncGodotProjectConfig.remove_project_settings()
+	#INFO: I realized that doing this defeats the purpose of using EditorSettings
+	# FuncGodotProjectConfig.remove_project_settings()
 
 ## Create the toolbar controls for [FuncGodotMap] instances in the editor
 func create_func_godot_map_control() -> Control:
--- a/addons/func_godot/src/func_godot_plugin.gd
+++ b/addons/func_godot/src/func_godot_plugin.gd
@@ -46,6 +46,9 @@ func _enter_tree() -> void:
 	add_control_to_container(EditorPlugin.CONTAINER_INSPECTOR_BOTTOM, func_godot_map_progress_bar)
 	
 	add_custom_type("FuncGodotMap", "Node3D", preload("res://addons/func_godot/src/map/func_godot_map.gd"), null)
+	
+	# Setup Editor Settings
+	FuncGodotLocalConfig.setup_editor_settings()
 
 func _exit_tree() -> void:
 	remove_custom_type("FuncGodotMap")
@@ -67,6 +70,8 @@ func _exit_tree() -> void:
 		remove_control_from_container(EditorPlugin.CONTAINER_SPATIAL_EDITOR_BOTTOM, func_godot_map_progress_bar)
 		func_godot_map_progress_bar.queue_free()
 		func_godot_map_progress_bar = null
+	
+	FuncGodotLocalConfig.remove_editor_settings()
 
 ## Create the toolbar controls for [FuncGodotMap] instances in the editor
 func create_func_godot_map_control() -> Control:
--- a/addons/func_godot/func_godot_local_config.tres
+++ b/addons/func_godot/func_godot_local_config.tres
@@ -1,7 +1,0 @@
-[gd_resource type="Resource" script_class="FuncGodotLocalConfig" load_steps=2 format=3 uid="uid://bqjt7nyekxgog"]
-
-[ext_resource type="Script" path="res://addons/func_godot/src/util/func_godot_local_config.gd" id="1_g8kqj"]
-
-[resource]
-script = ExtResource("1_g8kqj")
-export_func_godot_settings = false
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -31,7 +31,7 @@ signal unwrap_uv2_complete()
 var _map_file_internal: String = ""
 
 ## Map settings resource that defines map build scale, textures location, and more.
-@export var map_settings: FuncGodotMapSettings = load("res://addons/func_godot/func_godot_default_map_settings.tres")
+@export var map_settings: FuncGodotMapSettings = FuncGodotProjectConfig.get_setting(FuncGodotProjectConfig.PROPERTY.DEFAULT_MAP_CONFIG)
 
 @export_category("Build")
 ## If true, print profiling data before and after each build step.
--- a/addons/func_godot/src/util/func_godot_local_config.gd
+++ b/addons/func_godot/src/util/func_godot_local_config.gd
@@ -4,7 +4,6 @@
 ## DO NOT CREATE A NEW RESOURCE! This resource works by saving a configuration file to your game's *user://* folder and pulling the properties from that config file rather than this resource.
 ## Use the premade `addons/func_godot/func_godot_local_config.tres` instead.
 class_name FuncGodotLocalConfig
-extends Resource
 
 enum PROPERTY {
 	FGD_OUTPUT_FOLDER,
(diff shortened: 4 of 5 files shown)


addons/func_godot/src/func_godot_plugin.gd
```diff
@@ -67,6 +74,13 @@ func _exit_tree() -> void:
 		remove_control_from_container(EditorPlugin.CONTAINER_SPATIAL_EDITOR_BOTTOM, func_godot_map_progress_bar)
 		func_godot_map_progress_bar.queue_free()
 		func_godot_map_progress_bar = null
+	
+	# Cleanup Editor Settings
+	FuncGodotLocalConfig.remove_editor_settings()
+	
+	# Cleanup Project Settings
+	#INFO: I realized that doing this defeats the purpose of using EditorSettings
+	# FuncGodotProjectConfig.remove_project_settings()
```
Pspritechologist, line 83:
```suggestion
	
	# Cleanup Editor Settings
	#INFO: I realized that doing this defeats the purpose of using EditorSettings
	# FuncGodotLocalConfig.remove_editor_settings()
	
	# Cleanup Project Settings
	FuncGodotProjectConfig.remove_project_settings()
```

RhapsodyInGeek:
We'd already explored these options extensively back when it was first developed for Qodot (largely for use with the *QodotFGDModelPointClass* implementation) and they each had their own problems that led to us utilizing the *QodotProjectConfig* that became the current *FuncGodotLocalConfig* resource. We found it to be the most reliable, least cross-contaminating, and most source control friendly solution for what we were trying to accomplish. The biggest problems we had were ProjectSettings end up being shared across machines and EditorSettings end up shared across projects, when we needed to save settings on a project by project basis that did not get shared across machines.

You can see the discussion between myself and Qodot contributor gwenvis here:
https://github.com/QodotPlugin/Qodot/pull/95#issuecomment-1771366876

Pspritechologist:
> We'd already explored these options extensively back when it was first developed for Qodot (largely for use with the _QodotFGDModelPointClass_ implementation) and they each had their own problems that led to us utilizing the _QodotProjectConfig_ that became the current _FuncGodotLocalConfig_ resource. We found it to be the most reliable, least cross-contaminating, and most source control friendly solution for what we were trying to accomplish. The biggest problems we had were ProjectSettings end up being shared across machines and EditorSettings end up shared across projects, when we needed to save settings on a project by project basis that did not get shared across machines.
> 
> You can see the discussion between myself and Qodot contributor gwenvis here: [QodotPlugin/Qodot#95 (comment)](https://github.com/QodotPlugin/Qodot/pull/95#issuecomment-1771366876)

I thought about the per-user per-project thing a bit ago in regards to EditorSettings, and realized this solution doesn't maintain that, but I have to say I'm confused about the utility of it. In what situation would I *not* want my map editor config to persist across projects? Requiring I set it for each project I make is a large hassle when the data is identical. 

As for ProjectSettings, it maintains parity as the data is already version controlled. The difference is the user is no longer reliant on Project-wide config bound to arbitrary files, or worse bound to files within the addons directory where users shouldn't really be modifying.

RhapsodyInGeek:
> In what situation would I not want my map editor config to persist across projects? Requiring I set it for each project I make is a large hassle when the data is identical.

image

The ones I highlighted in green vary from project to project because they don't point to the TrenchBroom or NetRadiant folders, they point to the _game configuration folders_ within. The FGD folder is also a local directory path, which may or may not be within the TrenchBroom or NetRadiant folder (maybe it's pointed to your local JACK folder, which has its own setup too).

The data isn't identical from project to project unless you're reusing the same game configuration. Someone else may have more than one project with a different set of FGDs for each one, which require different folder locations.

Two settings I could see having cases made for them to be put in Project Settings would be `Game Path Models Folder` and `Default Inverse Scale Factor`, but they're not actually part of the build process: they're intended for FGD configuration settings, currently only used by the *FuncGodotFGDModelPointClass*. The Models folder is a relative location to the `Map Editor Game Path` where it saves the resulting GLB and points to it in the FGD. Default Inverse Scale is used to scale models when A) the target map editor is _NOT_ TrenchBroom, and B) when the entity definition does not have a scale expression applied... and that's it. 

That all said, while they could be put into Project Settings, we kept them with the other Local Config settings because it's overall more convenient to keep them together.

> As for ProjectSettings, it maintains parity as the data is already version controlled.

We don't want the data to be version controlled. That's the whole point of the Local Config JSON.

> The difference is the user is no longer reliant on Project-wide config bound to arbitrary files, or worse bound to files within the addons directory where users shouldn't really be modifying.

The user is still reliant on a Project-wide config bound to an arbitrary file; the difference is just the location of where you set them in the editor and where they're saved on disk. The user isn't actually modifying any resources when using the `func_godot_local_config.tres`, it's really just an interface using Resource system. It's designed specifically so that even when you update FuncGodot, it doesn't override your saved local configuration. The manual is pretty clear on this.

I know it may look and feel cleaner to put them into Godot's Project and Editor Settings, but we chose not to use them for very specific reasons.

Pspritechologist:
Gonna message you in the Discord to keep this moving a little faster

RhapsodyInGeek:
I prefer to have the discussion here tbh, so people can have a better record of the conversations that take place over development on the plugin. I'm also not likely to have the discussion move any faster by you throwing it into Discord, as when I'm good to have these kinds of PR development discussions I check the git channel to see if folks have responded.

We've had complaints during Qodot of plugin changes / additions / etc... being discussed on the Discord but not on GitHub, which I do agree with. On the Discord it's easy to see if a discussion is occurring on GitHub; the reverse is not true.

Pspritechologist:
Alright. For reference then, the message is here: https://discord.com/channels/1184354836762873876/1184936453113917450/1240141165270011915

The important bits are mostly be clarifying that ProjectSettings doesn't house anything that previously *wasn't* version controlled, and that there are tons of options to get the same functionality while making the interface quite a bit more intuitive, and cleanly integrated with Godot.
Also a lot of me gushing about both Godot and FuncGodot, clarifying the reason I consider this important. :P

RhapsodyInGeek:
From your messages on Discord (so folks don't have to jump around):

> So, I think you may have misunderstood me a bit. The only thing I'm moving to ProjectSettings is the default configuration used by map Nodes, allowing you to set defaults on a per-project basis, something which is A. Already local to the Project and B. Requires modifying files within the addons directory to set currently
The local config stuff is in EditorSettings 
As for the editor config stuff, perhaps the Game Config Folder could simply point to the directory that game data is placed in, and then create a folder named after the current project instead? Or to keep what you want, this functionality could be moved to a Tool accessed through the toolbar at the top as other addons do.
I just worry that using awkward solutions like creating a psuedo-interface inside a single-use Resource makes the addon feel less intuitive to use.
Trying to integrate the functionality as closely to the base editor as possible allows things to feel seamless and feel more like an extension to an editor rather than a cool tool someone made.
The main reason I push for this so hard is because I love Godot and what you can do with it, I love how open it is and how openly it exposes everything, the ability to take the editor and just crank it to be whatever you want is amazing and lets you legitimately expand it to be whatever you want without ever noticing it was modified in the first place
>
> In that same vein, I think FuncGodot is awesome, and I think it does something really important- allows Godot projects to take advantage of the amazing work being put into other tools- like TrenchBroom- to cover areas that the base Editor doesn't itself. Godot offers so much functionality, but making maps is not part of that, it frankly sucks at it. Using TB was my first time using a proper 'map editor' and I have to say, I see why people don't just do this in Blender :P
I have the utmost respect for this tool and what it allows, and when I started using it myself and saw some of the much more surface level areas of it lacking some polish that makes it feel more like a Godot-first extension, I thought I could probably suggest some things towards tuning it. It's not like I'll be contributing to the underlying mechanics of it any time soon :P 
I'm of the opinion that the end user should never need to open the addon's directory to access core functionality, even if it doesn't involve modifying files there. It doesn't feel great to parse the directory structure to configure things, and you need to remember where arbitrary files are to update them. I could make a simple 'tool' script to provide the same functionality, the local config options could be moved to Editor Settings (note- including the export if that's what you want), there's a ton of options I would be much more than happy to contribute if I can help in making this tool feel cleaner to use.
I know you have the concern of wasted time, but please don't worry about my time for me. I'm more than capable of recognizing what I want to do and what I should do and I understand the potential results of my actions and am more than able to handle it. 

Editor Settings get shared across projects even with FuncGodot not installed in them, and I don't mean just the values you input: the actual settings themselves seem to copy over as well. They persist even when you delete the plugin. We chose not to utilize them because of this.

This repeated suggestion that users should not explore the addons folder is also not really something I consider either a problem or a desire. FuncGodot is a complicated enough plugin that no matter what you need to follow the manual: it's not ever going to be a simple plug'n'play doing anything more than building worldspawn and func_detail. Additionally, unlike Qodot, users are encouraged to use the default resources that come with it in their own projects: the basic _func*_ classes, the TrenchBroom tags, the NetRadiant shaders... The plugin directory structure was reworked from Qodot so that you didn't have to dig too hard, everything is as obvious as we could make it.

That includes the ease of finding the `func_godot_local_config.tres`: it's literally in the root func_godot folder since it's one of the most important files in the plugin.

The only consideration I'd honestly consider making would be if the Project Settings were able to be set up to use the JSON file in `user://` exactly like the current resource interface setup. It'd have to be proven not to affect version control though, by pulling from the JSON config at project load and only updating it when an "update config" button is pushed. This would satisfy the original intent to have these settings in Project Settings without breaking compatibility with existing projects or polluting other unrelated projects.

Adding `default_map_settings` to the plugin is a good addition but it should be its own separate PR if it's not going to be related to Local Config. I'd wait on making a PR for it though, until we figure out if there is going to be a change to Local Config.

I also like the addition of the "default" values in the Local Config replacing the type check. This would also be better under its own PR.

In general, any feature update PRs should ideally stick to a single focus, with any other additions in the same PR being to serve that focus directly. It makes it easier to accept smaller fixes and features more quickly like the ones compiled into this much larger PR (and it is a large one insofar as its impact on how both the internals and interface of the plugin work).

Pspritechologist:
> Editor Settings get shared across projects even with FuncGodot not installed in them, and I don't mean just the values you input: the actual settings themselves seem to copy over as well. They persist even when you delete the plugin. We chose not to utilize them because of this.

I consider this a huge plus, it means settings are consistent. They're clearly placed and easy to remove for anyone who wants to. This is a non-issue to me, and part of Godot's great modularity. 

> This repeated suggestion that users should not explore the addons folder is also not really something I consider either a problem or a desire. FuncGodot is a complicated enough plugin that no matter what you need to follow the manual: it's not ever going to be a simple plug'n'play doing anything more than building worldspawn and func_detail. Additionally, unlike Qodot, users are encouraged to use the default resources that come with it in their own projects: the basic _func*_ classes, the TrenchBroom tags, the NetRadiant shaders... The plugin directory structure was reworked from Qodot so that you didn't have to dig too hard, everything is as obvious as we could make it.

I consider modifying addons a big issue for a number of reasons- The biggest being that any changes made in there completely breaks environment setup; I use a one of several tools to manage Godot plugins without ever committing them (only committing a list of plugins), this is extremely useful for a number of reasons but it totally disallows altering files since the entire directory is ignored.
As a side note, it allows client-side plugins without committing them.
(I am aware the local config isn't affected by this.)

> The only consideration I'd honestly consider making would be if the Project Settings were able to be set up to use the JSON file in `user://` exactly like the current resource interface setup. It'd have to be proven not to affect version control though, by pulling from the JSON config at project load and only updating it when an "update config" button is pushed. This would satisfy the original intent to have these settings in Project Settings without breaking compatibility with existing projects or polluting other unrelated projects.

This could be done quite easily.
What about at least changing the directory to be `get_project_settings_dir`, where project-specific editor settings are supposed to live?

> In general, any feature update PRs should ideally stick to a single focus, with any other additions in the same PR being to serve that focus directly. It makes it easier to accept smaller fixes and features more quickly like the ones compiled into this much larger PR (and it is a large one insofar as its impact on how both the internals and interface of the plugin work).

The purpose of this PR was to gauge interest in general, so I'm pretty happy with the results :P

RhapsodyInGeek:
> I consider this a huge plus, it means settings are consistent.

Literally just explained why we don't want this. These won't get merged as Editor Settings.

> I consider modifying addons a big issue for a number of reasons...

We don't tell people to _modify_ the resources - we actively _warn against modifying them_. We explain why the Local Config is the only one you can modify, but we also have declared the default resources as safe to ***use*** (not modify) since they won't be changing in their purpose and so won't break existing projects. If people want to modify them we always explain that they need to copy the files outside the addons folder. This is a non-issue for how this plugin is instructed to be used.

> What about at least changing the directory to be get_project_settings_dir, where project-specific editor settings are supposed to live?

Needlessly project breaking. While you could add conditionals to fall back to the JSON in `user://` with the first location to check being the project settings directory, there's honestly no practical reason to make the code more complicated than it needs to be.

> The purpose of this PR was to gauge interest in general, so I'm pretty happy with the results :P

It's fine to create discussions to gauge interest, but I'd still ask that you keep PRs focused more tightly.

PR #17, opened 2024-05-18
Implement support for ORMMaterial3D

meanwhile0:
Allows you to assign an ORMMaterial3D as the default material to use ORM maps in place of standalone AO, metallic and roughness maps.


RhapsodyInGeek:
The way you describe this PR (changing a default behavior for material generation) doesn't quite match the actual contents of it, but the actual content of it (adding optional support for ORM without disrupting current workflows) is actually what I would've wanted out of a PR.

Simple non-invasive and easily parsed feature addition. Looks good to me! Thanks for the contribution!

PR #18, opened 2024-05-21
Add export variables for default build XML

sinewavey:
I was tired of re-exporting the NRC config, and erasing my default build menu in the process.

Sure, I could just save a copy somewhere and replace it each time (no thanks) or use the desired file as an export variable -- but it's really just a few key/string value combos that can be written in a few arrays.

Here's a rough patch to include this, or at least allow the data to be retained export to export. Also, there was a minor typo in the print statements that was corrected.

As always, I don't expect this to be without some change or feedback.

### DIFF
--- a/addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
+++ b/addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
@@ -49,6 +49,12 @@ extends Resource
 ## Skip texture path that gets applied to caulk and nodrawnonsolid shaders.
 @export var skip_texture: String = "textures/special/skip"
 
+## Optional variables to include in the build profile.
+@export var build_xml_variables: Array[Array]
+
+## Optional commands to include in the build profile.
+@export var build_xml_commands: Array[Array]
+
 ## Generates completed text for a .shader file.
 func build_shader_text() -> String:
 	var shader_text: String = ""
@@ -66,23 +72,23 @@ func build_gamepack_text() -> String:
 		texturetypes_str += texture_type
 		if texture_type != texture_types[-1]:
 			texturetypes_str += " "
-	
+
 	var modeltypes_str: String = ""
 	for model_type in model_types:
 		modeltypes_str += model_type
 		if model_type != model_types[-1]:
 			modeltypes_str += " "
-	
+
 	var soundtypes_str: String = ""
 	for sound_type in sound_types:
 		soundtypes_str += sound_type
 		if sound_type != sound_types[-1]:
 			soundtypes_str += " "
-	
+
 	var gamepack_text: String = """<?xml version="1.0"?>
 <game
   type="q3"
-  index="1" 
+  index="1"
   name="%s"
   enginepath_win32="C:/%s/"
   engine_win32="%s.exe"
@@ -113,7 +119,7 @@ func build_gamepack_text() -> String:
   common_shaders_dir="common/"
 />
 """
-	
+
 	return gamepack_text % [
 		game_name,
 		game_name,
@@ -138,22 +144,22 @@ func do_export_file() -> void:
 	if (FuncGodotLocalConfig.get_setting(FuncGodotLocalConfig.PROPERTY.MAP_EDITOR_GAME_PATH) as String).is_empty():
 		printerr("Skipping export: Map Editor Game Path not set in Project Configuration")
 		return
-	
+
 	var gamepacks_folder: String = FuncGodotLocalConfig.get_setting(FuncGodotLocalConfig.PROPERTY.NETRADIANT_CUSTOM_GAMEPACKS_FOLDER) as String
 	if gamepacks_folder.is_empty():
 		printerr("Skipping export: No NetRadiant Custom gamepacks folder")
 		return
-	
+
 	# Make sure FGD file is set
 	if !fgd_file:
 		printerr("Skipping export: No FGD file")
 		return
-	
+
 	# Make sure we're actually in the NetRadiant Custom gamepacks folder
 	if DirAccess.open(gamepacks_folder + "/games") == null:
 		printerr("Skipping export: No \'games\' folder. Is this the NetRadiant Custom gamepacks folder?")
 		return
-	
+
 	# Create gamepack folders in case they do not exist
 	var gamepack_dir_paths: Array = [
 		gamepacks_folder + "/" + gamepack_name + ".game",
@@ -161,7 +167,7 @@ func do_export_file() -> void:
 		gamepacks_folder + "/" + gamepack_name + ".game/scripts"
 	]
 	var err: Error
-	
+
 	for path in gamepack_dir_paths:
 		if DirAccess.open(path) == null:
 			print("Couldn't open " + path + ", creating...")
@@ -169,10 +175,10 @@ func do_export_file() -> void:
 			if err != OK:
 				printerr("Skipping export: Failed to create directory")
 				return
-	
+
 	var target_file_path: String
 	var file: FileAccess
-	
+
 	# .gamepack
 	target_file_path = gamepacks_folder + "/games/" + gamepack_name + ".game"
 	print("Exporting NetRadiant Custom Gamepack to ", target_file_path)
@@ -182,7 +188,7 @@ func do_export_file() -> void:
 		file.close()
 	else:
 		printerr("Error: Could not modify " + target_file_path)
-	
+
 	# .shader
 	target_file_path = gamepacks_folder + "/" + gamepack_name + ".game/scripts/" + gamepack_name + ".shader"
 	print("Exporting NetRadiant Custom Shader to ", target_file_path)
@@ -192,27 +198,39 @@ func do_export_file() -> void:
 		file.close()
 	else:
 		printerr("Error: Could not modify " + target_file_path)
-	
+
 	# shaderlist.txt
 	target_file_path = gamepacks_folder + "/" + gamepack_name + ".game/scripts/shaderlist.txt"
-	print("Exporting NetRadiant Custom Default Buld Menu to ", target_file_path)
+	print("Exporting NetRadiant Custom Default Build Menu to ", target_file_path)
 	file = FileAccess.open(target_file_path, FileAccess.WRITE)
 	if file != null:
 		file.store_string(gamepack_name)
 		file.close()
 	else:
 		printerr("Error: Could not modify " + target_file_path)
-	
+
 	# default_build_menu.xml
 	target_file_path = gamepacks_folder + "/" + gamepack_name + ".game/default_build_menu.xml"
-	print("Exporting NetRadiant Custom Default Buld Menu to ", target_file_path)
+	print("Exporting NetRadiant Custom Default Build Menu to ", target_file_path)
 	file = FileAccess.open(target_file_path, FileAccess.WRITE)
 	if file != null:
-		file.store_string("<?xml version=\"1.0\"?><project version=\"2.0\"></project>")
+		file.store_string("<?xml version=\"1.0\"?>\n<project version=\"2.0\">\n")
+
+		for v in build_xml_variables:
+			if !(v is Array) || v.size() != 2:
+				continue
+			file.store_string('	<var name="%s">%s</var>\n' % [v[0], v[1]])
+
+		for c in build_xml_commands:
+			if !(c is Array) || c.size() != 2:
+				continue
+			file.store_string('	<build name="%s">\n		<command>%s</command>\n	</build>\n' % [c[0], c[1]])
+
+		file.store_string("</project>")
 		file.close()
 	else:
 		printerr("Error: Could not modify " + target_file_path)
-	
+
 	# FGD
 	var export_fgd : FuncGodotFGDFile = fgd_file.duplicate()
 	export_fgd.do_export_file(true, gamepacks_folder + "/" + gamepack_name + ".game/" + base_game_path)


addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -66,23 +72,23 @@ func build_gamepack_text() -> String:
 		texturetypes_str += texture_type
 		if texture_type != texture_types[-1]:
 			texturetypes_str += " "
-	
+
```
RhapsodyInGeek, line 75: It's really hard to track the actual changes when you've (I assume accidentally?) deleted every indentation that doesn't have any code on the same line. Is there any chance you could undo that and avoid it in the future?

sinewavey: Oh shoot, that's my Godot editor auto clearing whitespace. Will fix this for the future + this commit as well

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -49,6 +49,12 @@ extends Resource
 ## Skip texture path that gets applied to caulk and nodrawnonsolid shaders.
 @export var skip_texture: String = "textures/special/skip"
 
+## Optional variables to include in the build profile.
+@export var build_xml_variables: Array[Array]
```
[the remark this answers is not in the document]
sinewavey: I was split between which one to use anyway. Not a problem.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -49,6 +49,12 @@ extends Resource
...
+
+## Optional commands to include in the build profile.
+@export var build_xml_commands: Array[Array]
```
RhapsodyInGeek, line 56: Looking at how build options and commands are constructed I think it'd make sense to also make these Dictionaries, but they'd be treated as `Dictionary<String><Array>`, as there can be multiple commands in a build option. Would make them easier to parse, more intuitive.

RhapsodyInGeek:
I totally forgot we generated the default_build_menu.xml, even though it makes sense that we do, why would NRC autogenerate that for us when it doesn't autogenerate anything except my misery?

sinewavey:
> I totally forgot we generated the default_build_menu.xml, even though it makes sense that we do, why would NRC autogenerate that for us when it doesn't autogenerate anything except my misery?

Lol, yep. Time to fix that other docs page again lol

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
RhapsodyInGeek: The main reason I don't clean up white space indentations in GDScript is that it makes it easier to track what lines of code belong to which code blocks when they're spaced apart with line breaks when in the Godot's script editor. Makes it easier to parse when you're dealing with extremely long scripts (like func_godot_map.gd). I'm honestly not sure of any good reason to clean up white space indentations to be honest, since there isn't a measurable impact on performance.

That doesn't mean white space _spaces_ shouldn't be cleaned up, but that's also not really a rampant problem as far as I've seen.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -49,6 +49,12 @@ extends Resource
 ## Skip texture path that gets applied to caulk and nodrawnonsolid shaders.
 @export var skip_texture: String = "textures/special/skip"
 
+## Optional variables to include in the build profile.
+@export var build_xml_variables: Dictionary
```
RhapsodyInGeek, line 53: Should be `default_build_menu_variables`, since we don't actually use the term `build_xml` anywhere and it's too much of a shorthand. Exposed variables should always be fully descriptive and easily associated with other processes and exports.

Can we also get a more expository description on what these variables are for? Not quite a tutorial, but since NRC is so poorly documented it might be helpful to let NRC users know what this is for.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -208,10 +214,15 @@ func do_export_file() -> void:
...
-		file.close()
-	else:
-		printerr("Error: Could not modify " + target_file_path)
+		file.store_string("<?xml version=\"1.0\"?>\n<project version=\"2.0\">\n")
+		
+		for key in build_xml_variables.keys():
+			file.store_string('\t<var name="%s">%s</var>\n' % [key, build_xml_variables[key]])
+			
+		for key in build_xml_commands.keys():
+			file.store_string('\t<build name="%s">\n\t\t<command>%s</command>\n\t</build>\n' % [key, build_xml_commands[key]])
```
RhapsodyInGeek, line 223: This should be a for loop that loops through the build commands array, since you can have multiple commands in one build option. Even though we haven't found a use for multiple in our own projects, we should still leave the theoretical possibility open since it's easy enough to implement it.

Better still would be an option to support both, since sometimes you only want the one option:
```gdscript
	for key in default_build_menu_commands.keys():
		file.store_string("'\t<build name=" + key + ">\n")
		if default_build_menu_commands[key] is String:
			file.store_string("\t\t<command>" + default_build_commands[key] + "</command>\n"
		elif default_build_menu_commands[key] is Array:
			var commands: Array = default_build_menu_commands[key]
			for command in commands:
				if command is String:
					file.store_string("\t\t<command>" + (command as String) + "</command>\n"
				else:
					push_error("ERROR: A command in build option \"" + key + "\" commands Array is not a String")
		else:
			push_error("ERROR: Build option \"" + key + "\" command is not a String or Array of Strings")
```

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -208,10 +214,15 @@ func do_export_file() -> void:
...
 	file = FileAccess.open(target_file_path, FileAccess.WRITE)
 	if file != null:
-		file.store_string("<?xml version=\"1.0\"?><project version=\"2.0\"></project>")
-		file.close()
-	else:
-		printerr("Error: Could not modify " + target_file_path)
+		file.store_string("<?xml version=\"1.0\"?>\n<project version=\"2.0\">\n")
+		
+		for key in build_xml_variables.keys():
+			file.store_string('\t<var name="%s">%s</var>\n' % [key, build_xml_variables[key]])
```
RhapsodyInGeek, line 220: Should do a safety check to make sure we're using a string here.

```gdscript
		for key in default_build_menu_variables.keys():
			if default_build_menu_variables[key] is String:
				file.store_string('\t<var name="%s">%s</var>\n' % [key, build_xml_variables[key]])
			else:
				push_error("ERROR: Build menu variable \"" + key + "\" is not a String")

Comment (sinewavey) [reply addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd:220]: Sure - %s casts to string I thought but it should still probably be caught anyway.

Comment (RhapsodyInGeek) [reply addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd]: AFAIK not all types can be cast to String though. I know there's a special `str(value)` in GDScript but does `%s` use that to convert it to String or does it just do the equivalent of `value as String`?

File: addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd [hunk 7]
```diff
@@ -49,6 +49,19 @@ extends Resource
 ## Skip texture path that gets applied to caulk and nodrawnonsolid shaders.
 @export var skip_texture: String = "textures/special/skip"
 
+## Variables to include in the exported gamepack's [code]default_build_menu.xml[/code].[br][br]
+## Each [String] key defines a variable name, and its corresponding [String] value as the literal command-line string to execute in place of this variable identifier[br][br]
+## Entries may be referred to by key in [member default_build_commands] values.
+@export var default_build_variables: Dictionary
```
RhapsodyInGeek, line 55: `default_build_menu_variables` specifically because these are added to the default_build_menu file.

Description is good though!

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -49,6 +49,19 @@ extends Resource
...
+
+## Commands to include in the exported gamepack's [code]default_build_menu.xml[/code].[br][br]
+## Keys, specified as a [String], define the command name as you want it to appear in Radiant.[br][br]
+## Values represent steps taken within the build command.[br][br]They may be either a [String] or an
```
RhapsodyInGeek, line 59: Keys are build options, not commands. Values are commands, which are equivalent to steps but listed as commands in NRC, both in the way that the files are formatted and how they show up in NRC. From the Half-Life default_build_menu.xml:
```xml
<build name="CSG, BSP, VIS -fast">
       <command>[csg] "[MapFile]"</command>
       <command>[bsp] "[MapFile]"</command>
       <command>[vis] -fast "[MapFile]"</command>
  </build>
```
So yeah, build options > commands, not commands > steps. I'm sorry if I sound pedantic, but accuracy is important to me for the docs.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -49,6 +49,19 @@ extends Resource
...
+## [Array][[String]] and will be used as the full command-line text issued by each step [i]within[/i]
```
RhapsodyInGeek, line 60: I'd have to double check, but I don't believe you can specify type for Arrays in Dictionaries made in the inspector. If that's the case, we'd have to rewrite this to more accurately describe it as 
```
## This may be either a [String] or an [Array] with [String] elements that will be used as the full command-line text issued by each command within its associated build option key.
```

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -49,6 +49,19 @@ extends Resource
...
+## its associated build command key. [br][br]They may reference entries in [member default_build_variables]
+## by using brackets: [code][variable key name][/code]
+@export var default_build_commands: Dictionary
```
RhapsodyInGeek, line 63: As above, should be `default_build_menu_commands` as it gets applied to the default_build_menu.xml.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -207,11 +220,50 @@ func do_export_file() -> void:
...
+	
 	if file != null:
-		file.store_string("<?xml version=\"1.0\"?><project version=\"2.0\"></project>")
-		file.close()
-	else:
-		printerr("Error: Could not modify " + target_file_path)
+		file.store_string("<?xml version=\"1.0\"?>\n<project version=\"2.0\">\n")
+		
+		for key in default_build_variables.keys():
+			if key is String:
```
RhapsodyInGeek, line 228: Good call.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -207,11 +220,50 @@ func do_export_file() -> void:
...
+							file.store_string('\t\t<command>%s</command>\n' % step)
+						else:
+							push_error("Command '%s' has invalid step: %s with type: %s; should be: String" % [
+								key, step, type_string(typeof(step))
+								])	
+						
+					file.store_string('\t</build>\n')
+			
+			else:
+				push_error("Command '%s' is an invalid type: %s; should be: String" % [
```
RhapsodyInGeek, line 262: `Command` should be `Build option`, as reasoned above. Don't use single quotes for referencing names, use double quotes (as should be the standard throughout the rest of the plugin).

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -207,11 +220,50 @@ func do_export_file() -> void:
...
+				
+				if default_build_commands[key] is String:
+					file.store_string('\t\t<command>%s</command>\n\t</build>\n' % default_build_commands[key])
+				
+				elif default_build_commands[key] is Array:
+					for step in default_build_commands[key]:
+						if step is String:
+							file.store_string('\t\t<command>%s</command>\n' % step)
+						else:
+							push_error("Command '%s' has invalid step: %s with type: %s; should be: String" % [
```
RhapsodyInGeek, line 255: `Command` should be `Build option`. Don't use single quotes for referencing names, use double quotes (as should be the standard throughout the rest of the plugin). `invalid step` should be `invalid command`. 

I like the addition of the invalid type to give a fuller report of what is wrong.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
```diff
@@ -207,11 +220,50 @@ func do_export_file() -> void:
...
+			
+		for key in default_build_commands.keys():
+			if key is String:
+				file.store_string('\t<build name="%s">\n' % key)
+				
+				if default_build_commands[key] is String:
+					file.store_string('\t\t<command>%s</command>\n\t</build>\n' % default_build_commands[key])
+				
+				elif default_build_commands[key] is Array:
+					for step in default_build_commands[key]:
```
RhapsodyInGeek, line 251: `step` to `command` for code clarity, for the reasons mentioned above.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
sinewavey: Correct, and while that was what I implied - better to outright state.

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
RhapsodyInGeek: It's not really exhausting for me; playing "Editor" isn't difficult. It's also interesting to open up NRC's guts a little, a small change of pace. I mostly get worried about offending by "a thousand cuts".

addons/func_godot/src/netradiant_custom/netradiant_custom_gamepack_config.gd
sinewavey: I'm not sure, actually, so that's a great point. Will fix

RhapsodyInGeek:
@sinewavey just wanted to make sure this was all good to go before merging? I read it over again and it looks good to me, but still wanted to double check with you.

sinewavey:
Sounds good to me

PR #19, opened 2024-06-03
Fix outdated build_def_text() signature on ModelPointClass

DeerTears:
Fixes an issue where Godot would not allow ModelPointClass or PointClass resources to be loaded into the project.

When the build_def_text() function was updated to prefer an enum of editors instead of a bool, this wasn't updated in ModelPointClass, which uses `super()`, which prevented PointClass from loading as well. This prevented any PointClass resources or any classes extending it from being added to new projects using Godot 4.2.2 stable and this latest github repository.

New PointClass resources were not available in the resource list, new ModelPointClass resources were cancelled before giving the user the dialog box to name their new resource.

## New Behaviour
PointClass and ModelPointClass resources can be created in the project, saved, and re-opened.

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_model_point_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_model_point_class.gd
@@ -32,7 +32,7 @@ enum TargetMapEditor {
 		file.store_string('')
 		file.close()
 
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
Ahh, big miss on my part. Thanks for catching it.

Issue #15, opened 2024-05-09
Fix Phong shading to be per-mesh and not per-brush


RisingThumb:
> Proper Phong Shading
> One of the features in FuncGodot is using the _phong and _phong_angle key value pairs to modify generated meshes vertex normals, to allow for smoother shading on brush entities. Currently the implementation only does this on a per brush basis, when it should be done on a per mesh basis. This issue was around in Qodot and has been carried over to FuncGodot. Hopefully we can get this sorted out in the near future

Off the back of the discussion page https://github.com/orgs/func-godot/discussions/2 .

RhapsodyInGeek:
Gonna be looking into this again soon. Any other help on this would be greatly appreciated.

RhapsodyInGeek:
Started work on trying to solve this. Using this branch of the test_project repository:
https://github.com/func-godot/func_godot_test_project/tree/phong-testing

RhapsodyInGeek:
High level overview of the phong process given to me by ericw on the Quake Mapping Discord:

> I can give a high level overview, there are a couple steps:
> - it runs on the output of qbsp, so hidden/exterior faces are deleted, everything is welded together
> - first, it's marking edges of the mesh as 'hard' or 'smooth' (same idea as in Blender) based on the _phong_angle (default is 89) and looking at the angle between neighbouring faces compared with the phong angle. So e.g. if you have a cube, the angle between the top face's normal and the sides' normals are all 90 degrees, so those 4 edges around the top face would get marked as "hard"
> - once edges are marked as hard or smooth we calculate vertex-per-face normals. Hard edges just use the face normal, and smoothed edges we're calculating a blended normal by blending together the face normals of the faces sharing the smoothed edge. I forget the exact details here but it's weighted by angle/face area

I believe the answer is going to be in GeoGenerator's `run()` method. Our main problem is that phong and normal calculations occur on a per brush basis.

My current idea is to remove the phong calculations from the brush generation section and create another threaded loop after the entity's brushes are generated that goes through each brush's vertices and corrects their normals.

PR #23, opened 2024-07-20
Add option to bake texture names into convex collision shapes

the-argus:
Some users (including myself) have expressed the desire to get the texture on a colliding brush, for the purpose of playing different footstep sounds when walking on differently textured surfaces. This PR adds an option, `bake_texture_names`, for these users. It only works for solid classes with convex collision.

Unfortunately, the information is buried all the way in the convex collision shape resource, which does lead to a bit of finagling to get access to it. Here is the code I added to the bottom of `marsfrog.gd` in the example basic project to test this:
```gdscript
func _physics_process(delta: float):
	if Engine.is_editor_hint():
		return
	super._physics_process(delta)
	var floor_result := _floor_raycast()
	if not floor_result.is_empty():
		var collider := floor_result["collider"] as StaticBody3D
		if collider:
			var shape_id := floor_result["shape"] as int
			var shape := collider.shape_owner_get_owner(collider.shape_find_owner(shape_id)) as CollisionShape3D
			if shape:
				var shape_resource := shape.shape as FuncGodotConvexPolygonShape3D
				if shape_resource:
					print("floor texture: ", shape_resource.get_texture_of_collision(floor_result["normal"]))

func _floor_raycast() -> Dictionary:
	var query := PhysicsRayQueryParameters3D.create(global_position, global_position + Vector3.DOWN)
	query.collide_with_areas = false
	query.exclude = [self]
	# hardcoded, i know that the first bit is for map geo
	query.collision_mask = 1
	return get_world_3d().direct_space_state.intersect_ray(query)
```

Additionally, it requires that users create their own brush types for floor geometry, for example `func_detail_convex`, since the ones provided with func_godot use concave collisions.


RhapsodyInGeek:
While this is cool, it can't be merged without support for concave collision, which is important for performance (contrary to popular belief, concave is not automatically less performant than convex shapes due to how Godot optimizes collision) and flexibility. Without concave support from the get-go, this PR is a no-go.

Additionally, I have concerns about the increased build times having to loop through every single face again. It may not seem like a big deal on small maps, but you need to also test it on significantly larger maps. It's common practice for Quake mappers to share the map files - [Slipseer](https://www.slipseer.com/index.php) should have plenty to choose from.

In order to minimize loops and stalls you should be gathering this information in GeoGenerator or SurfaceGatherer, where they are already looping through brush faces and textures.

I like the idea, but it definitely doesn't work as-is for what FuncGodot needs.

RisingThumb:
If I understand correctly this PR does something similar to what this other PR does #8 . It may also be worth looking at the discussion that was had around it

the-argus:
Okay, added support for concave collision shapes. Using it requires the following change to the `if shape` clause of the code example given in the first comment:

```gdscript
			if shape:
				var shape_resource_convex := shape.shape as FuncGodotConvexPolygonShape3D
				var shape_resource_concave := shape.shape as FuncGodotConcavePolygonShape3D
				if shape_resource_convex:
					print("floor texture: ", shape_resource_convex.get_texture_of_collision(floor_result["normal"]))
				elif shape_resource_concave:
					print("floor texture: ", shape_resource_concave.get_texture_of_collision(floor_result["face_index"]))
```

I've also moved the generation of texture information into `FuncGodotSurfaceGatherer`. However:

> Additionally, I have concerns about the increased build times having to loop through every single face again. It may not seem like a big deal on small maps, but you need to also test it on significantly larger maps. It's common practice for Quake mappers to share the map files - [Slipseer](https://www.slipseer.com/index.php) should have plenty to choose from.

I have not tested this yet. I will do benchmarks (for speed and scene filesize) for some of these maps with and without my changes, however I'm not sure how helpful that would be, as the texture name info is off by default. I guess I will also benchmark it with every solid entity having it on by default? Are there any other benchmarks you'd like to optimize for?

Some issues I (might?) have with my implementation:
- I think I have made it substantially harder for new people to read the SurfaceGatherer code (my biggest concern)
- The process of getting the collision shape and calling the `get_texture_of_collision` function is quite tedious. I think its good to keep responsibility off of the plugin, but also it might be nice to add some `get_texture_raycast(raycast_result: Dictionary) -> String` function.
- Texture lookup on concave colliders could be sped up by doing a binary search, easy optimization but would need benchmarking, see above
- Generally, I think there are opportunities to sacrifice build time for better runtime performance. I think it might be the case that a lot of brushes will only have one texture applied besides clip and skip, maybe useful for convex collision lookups?
- I created some types in SurfaceGatherer, I think this makes sense since they only come from there, but theres also the pattern of putting all the type in `func_godot_map_data.gd`
- One of the types has to be converted to a dictionary so it can be serialized in the `@export` field. Originally I wrote it as a dictionary, so no conversion was needed, but I felt it was difficult to decipher what the data structures contained without reading the rather complicated `run()` function that generated them.

> If I understand correctly this PR does something similar to what this other PR does #8 . It may also be worth looking at the discussion that was had around it

My takeaway here is that there's a possibility that in the future we will get different physics materials based on texture/material. However it seems that this is a ways off, and also I am not in a position to do it, seems best left to core maintainers since it involves refactoring. My other takeaway is that I could use metadata instead of inheriting from collision shape resources. I am not sure there are an advantages/disadavantages either way, though. Also, it seems my choice to make baking texture info a flag in SolidClass was the right move, so thats good. Lmk if I missed anything.

RisingThumb:
> My takeaway here is that there's a possibility that in the future we will get different physics materials based on texture/material. However it seems that this is a ways off, and also I am not in a position to do it, seems best left to core maintainers since it involves refactoring. My other takeaway is that I could use metadata instead of inheriting from collision shape resources. I am not sure there are an advantages/disadavantages either way, though. Also, it seems my choice to make baking texture info a flag in SolidClass was the right move, so thats good. Lmk if I missed anything.

Have you tested if it works for a func_illusionary(a solid brush without any collisions)? It should be possible to get the list of textures for it too(which iirc means no collision shape node/resource is added or used for it).

```
var godot_normal := Vector3(face.plane_normal.y, face.plane_normal.z, face.plane_normal.x)
normal_to_texture_map[godot_normal] = map_data.textures[face.texture_idx].name
```
This code is flawed for concave collisions.
ConcaveIssue
To demonstrate this(in 2D), Line AB and line CD in the above image have the same normal, which means here you have a race condition of whichever face gets assigned last is the one that normal considers to be the texture, even if it's incorrect. This is also the same issue as line BC and line DE. Some more information is needed to distinguish the two faces and textures(There's no reason it can't be 3 or more faces sharing the same normal).

Personally, I think this could be broken up into 2 separate problems. The first is getting the textures of a brush(which is what #8 set out to solve), and the 2nd is matching textures back up to a face in collisions. This latter point afaik, has no easy solution in Godot due to the separation of MeshInstance and Collisons(and afaik, Unity and Unreal also has the same issue, though the Source engine doesn't iirc).

RhapsodyInGeek:
@the-argus 
Concave and Convex texture look up shouldn't be different honestly. If they can both be done by face, they should both be done by face, since you should be using both in your projects, not one or the other. Doors, for example, would be AnimatableBody or CharacterBody, maybe RigidBody... all of which need to use convex collision. StaticBody meanwhile should be concave to take advantage of PhysicsServer optimizations.

You're looping through each face again here. This dictionary should be made in SurfaceGatherer where you get that information in the first place.
```gdscript
func populate_texture_info(index_ranges: Array[FuncGodotSurfaceGatherer.BrushTextureIndexRange]) -> void:
	_generated_texture_lookup.clear()
	# converting type -> serializable dict. could always be dict instead for better perf, worse readability
	for index_range in index_ranges:
		_generated_texture_lookup.append({
			texture_name = index_range.texture_name,
			start = index_range.start,
			end = index_range.end,
		})
```

> I think I have made it substantially harder for new people to read the SurfaceGatherer code (my biggest concern)

I didn't have trouble reading it.

> The process of getting the collision shape and calling the get_texture_of_collision function is quite tedious. I think its good to keep responsibility off of the plugin, but also it might be nice to add some get_texture_raycast(raycast_result: Dictionary) -> String function.

That's out of scope for FuncGodot.

> Generally, I think there are opportunities to sacrifice build time for better runtime performance.

We still need to properly implement phong #15 , and my performance tests on several solutions I tried were not promising. Not interested in exacerbating issues when they don't have to be.

> I created some types in SurfaceGatherer, I think this makes sense since they only come from there, but theres also the pattern of putting all the type in func_godot_map_data.gd

If all of Core's types are in MapData, then the new types should also go there.

> I think maybe the most appropriate thing would be to batch the texture info and corresponding collision shapes and then make another build step to perform the conversion.

I would rather we try to reduce build steps rather than increase them. Try to find appropriate pre-existing loops to insert these steps into. For Core, while we should do both if it's possible to do both, the preference is performance over elegance.

> My other takeaway is that I could use metadata instead of inheriting from collision shape resources. I am not sure there are an advantages/disadavantages either way, though.

Having the extended collision shape resources is helpful insofar as they provide a means for texture lookup information. That said, it also straddles the line of designing a developer's game for them, which we want to avoid if possible. My default position is to have the information provided, but the means to fetch it or manipulate it should be left in the hands of the game developer. There are cases where I can be persuaded away from this position, but they should be strong cases.

> Also, it seems my choice to make baking texture info a flag in SolidClass was the right move, so thats good.

Yes.

@RisingThumb 
> Have you tested if it works for a func_illusionary(a solid brush without any collisions)? It should be possible to get the list of textures for it too(which iirc means no collision shape node/resource is added or used for it).

You'll already be able to access the materials list from the MeshInstance3D directly. While I'm not against this, I'm also not convinced of the usefulness when the data is already available.

> Personally, I think this could be broken up into 2 separate problems. The first is getting the textures of a brush(which is what https://github.com/func-godot/func_godot_plugin/pull/8 set out to solve), and the 2nd is matching textures back up to a face in collisions. This latter point afaik, has no easy solution in Godot due to the separation of MeshInstance and Collisons(and afaik, Unity and Unreal also has the same issue, though the Source engine doesn't iirc).

While I don't disagree, the thing I like about what @the-argus has done here is combining the two without needing to use an additional loop. Would like to reduce loops if we can.

That said, @the-argus have you actually tested the concave collision detection with deleted collision faces? Do they still match up with the index lookups? I'm skeptical that they would, since you seem to include the filtered faces in your data structures.

the-argus:
> Have you tested if it works for a func_illusionary(a solid brush without any collisions)? It should be possible to get the list of textures for it too(which iirc means no collision shape node/resource is added or used for it).

With this solution there is no way to get the textures on a mesh by position in the world. All of this relys on godot's collision detection. One option to do this could be to make the illusory brush generate an Area3D, and check collisions with that, probably by raycast.

> ```
> var godot_normal := Vector3(face.plane_normal.y, face.plane_normal.z, face.plane_normal.x)
> normal_to_texture_map[godot_normal] = map_data.textures[face.texture_idx].name
> ```
> 
> This code is flawed for concave collisions.

Indeed, that is why that code only runs when convex colliders are used. Just to be sure, I built a similar shape to the one you provided an image of and tested it, both in convex and concave mode. Successfully distinguished between the textures on faces with the same normal! (this is thanks to the convex collision having to be split into multiple colliders to express the concave shape)

> Personally, I think this could be broken up into 2 separate problems. The first is getting the textures of a brush(which is what #8 set out to solve), and the 2nd is matching textures back up to a face in collisions. This latter point afaik, has no easy solution in Godot due to the separation of MeshInstance and Collisons(and afaik, Unity and Unreal also has the same issue, though the Source engine doesn't iirc).

I disagree. I think that there is no need to perform any texture -> collision matching step. When we make levels in trenchbroom, JACK, whatever, we use brushes, and so we define the level in terms of collisions _and_ mesh, together. The information is already there, so just propagate the information into a gdscript-readable format, from the .map file. That's my approach here at least, and its simple and seems to work.

From RhapsodyInGeek:
> Concave and Convex texture look up shouldn't be different honestly. If they can both be done by face, they should both be done by face, since you should be using both in your projects, not one or the other. Doors, for example, would be AnimatableBody or CharacterBody, maybe RigidBody... all of which need to use convex collision. StaticBody meanwhile should be concave to take advantage of PhysicsServer optimizations.

RhapsodyInGeek:
> Note: Returns a valid number only if the intersected shape is a [ConcavePolygonShape3D](https://docs.godotengine.org/en/stable/classes/class_concavepolygonshape3d.html#class-concavepolygonshape3d). Otherwise, -1 is returned.

That's a shame.

> One of my tests was to make a big entity of concave terrain terrain out of loads of triangle shaped brushes, and texture random faces all over clip/skip/sand0/sand1 etc. I tried it both with convex and concave, and everything seemed to work fine. Does that sound like it would test the case youre talking about?

Yes, particularly skip and clip since skip removes both collision and mesh faces, and clip removes mesh faces.

Also, changing my stance on the extended collision shape classes [due to a conversation had on the Discord](https://discord.com/channels/1184354836762873876/1184936453113917450/1264982405371133955):
> It’s also preferred not to require FuncGodot for Runtime stuff, keep it optional. That is one problem the collision shape classes you had did raise: it suddenly creates a dependency on the plug-in outside of the build process.
Easy to forget that once you build your map currently, you don’t actually need to keep FuncGodot installed. You can change the FuncGodotMap node into a Node3D and it should still just work.
Assuming your node classes themselves don’t reference FuncGodot.

Gonna go ahead and suggest these texture / collision dictionaries should be added to the Nodes' Metadata instead.

sinewavey:
I like the idea of node metadata here - it's not intrusive, and may be independent of requiring func_godot to interface with. On top of that, it 'frees up' a hard class requirement, which would allow it to go on many kinds of user implemented classes as well.

the-argus:
Okay, I have removed the extra scripts and I do agree its much nicer now! However, it does mean more needs to be documented as raw access to the data structures means more code fro the user to write. I think it's worth putting this script here, which I believe most users would need to reimplement if they use my branch in its current state:

```gdscript
class_name GetTextureUtil

static func get_texture_of_raycast(cast_result: Dictionary) -> String:
	if cast_result.is_empty():
		return ""

	var collider := cast_result["collider"] as StaticBody3D
	if not collider:
		return ""

	var shape_id := cast_result["shape"] as int
	var shape := collider.shape_owner_get_owner(collider.shape_find_owner(shape_id)) as CollisionShape3D
	if not shape:
		return ""

	if shape.shape is ConcavePolygonShape3D:
		return _get_tex_concave_collision(shape, cast_result["face_index"])
	elif shape.shape is ConvexPolygonShape3D:
		return _get_tex_convex_collision(shape, cast_result["normal"])
	return ""

static func _get_tex_concave_collision(shape_node: CollisionShape3D, face_index: int) -> String:
	# NOTE: face ranges are sorted, this could be binary search
	for range: Dictionary in shape_node.get_meta("func_godot_texture_index_ranges", []):
		var start := range["start"] as int
		var end := range["end"] as int
		if face_index >= start and face_index < end:
			return range["texture_name"] as String

	return ""

static func _get_tex_convex_collision(shape_node: CollisionShape3D, normal: Vector3) -> String:
	const FLOATING_POINT_ERROR_EPSILON := 0.0001
	var lookup: Dictionary = shape_node.get_meta("func_godot_normal_texture_lookup")
	if lookup.has(normal):
		return lookup[normal]

	# maybe we have it but floating point madness happened in physics,
	# given normal is just slightly different than actual
	for tex_normal in lookup.keys():
		if tex_normal.dot(normal) > 1 - FLOATING_POINT_ERROR_EPSILON:
			return lookup[tex_normal]

	return ""
```

I like that the user gets to see the data and process it, particularly for the convex solution which is not foolproof and potentially has  unpreditable performance characteristics.

RhapsodyInGeek:
Closing in favor of #60

Issue #26, opened 2024-08-19
Compatibility Issues with Older TrenchBroom Versions and GameConfig.cfg Export Format


Pesquisador-Stefano:
Hi everyone,

I'm currently working with a PC that has lower specs than most modern calculators. Unfortunately, I can't run the newer versions of TrenchBroom because my system doesn't support the latest OpenGL versions (I can only run Godot in compatibility mode). However, I can still run TrenchBroom versions older than 2021.1.

The problem I'm encountering is that the 'trenchbroom_game_config.gd' script exports the 'GameConfig.cfg' in a newer format, which isn't compatible with these older TrenchBroom versions. While I understand this isn't a common issue for most users, it would be helpful for those of us with older systems if there were an option to export in the older format as well.

I could manually create a GameConfig.cfg file, but this wouldn't improve usability. While I can still create independent maps, having compatibility with older versions would greatly enhance the workflow for anyone in a similar situation.

RhapsodyInGeek:
Unfortunately the Core Team is small, and trying to support older TrenchBroom versions is complex and, as you can imagine, not very high on our priority when most of our users tend to switch to the latest version. One of our core design rules is to only concern ourselves with the latest versions of the tools we want to be compatible with, since it's more than just TrenchBroom we have to pay attention to. We also have a lot of other projects we work on.

That said, FuncGodot is a fully open source project. If you or some other contributor can add backwards compatibility to the appropriate resources via a drop down menu that always defaults to the latest version, I'd be open to guiding that PR (though I'm quite busy and admittedly don't have this issue so I won't be hands on).

TrenchBroom does typically keep pretty good documentation on the GameConfig formats. Older versions of Qodot also have setups for older versions. I'd recommend researching those and the GameConfig files that come included with TrenchBroom 2021 heavily before trying to implement them.

Leaving this issue open for now to allow discussion and potential implementation. Will close if it becomes inactive or resolved.

RisingThumb:
The way this is implemented should consider future GameConfig changes too, and allowing it to be simple for the old gameconfig format to added.

Additionally if this is added, it'd be worthwhile noting somewhere(in the docs?) these gameconfig versions aren't actively supported by the core developers or widely used by people in general. I don't think Trenchbroom goes out of their way to make bug fixes for old versions either

RhapsodyInGeek:
[...]

PR #27, opened 2024-08-21
Added compatibility for Trenchbroom version 2021.1 

Pesquisador-Stefano:
Things changed (the commit description):
- Generated GameConfig.cfg now supports older versions of Trenchbroom
- Added compatibility for the game config version4 ~~in Trenchbroom version 2021.1~~
- ~~The brush_tags variable now has a default value (I removed it and was unaware it was necessary).~~

So... I think I managed to find a solution to #26, but the thing is:
1. I don't know anything about Trenchbroom; I've never used it before, and I'm starting to mess with it now.
2. The same thing can be said for FuncGodot, I still need to figure out how to add textures to ~~sh~~models.
3. I can't run/open Trenchbroom maps on newer versions.

Since I don't have the knowledge nor the machine to run Trenchbroom, it will take a long time to test things. So if it's not too much to ask, can someone test if these changes work? ~~(I doubt they do)~~

~~P.S: Is there a way i can remove some changes from this PR? i.e: ".gitignore"~~


RhapsodyInGeek:
I would avoid using text files or multiple files outside of the class. Instead I'd opt to keep everything contained within the TrenchbroomGameConfig resource script.

The way it's set up seems overdesigned, seems like it could break configs with team projects, and it's also using the TrenchBroom version numbers. Instead it should use the GameConfig Version numbers, as they don't always change when the editor version does.

Version should be determined with a `var game_config_version: int = 0`. I'd also recommend an enum instead to determine the GameConfig version numbers, in ascending order. Something like:
```gdscript
enum GameConfigVersion {
    Latest,
    Version4,
    Version8
}
```
Then the user would select the `game_config_version` in a dropdown.

Each version should have its own dedicated function that returns the GameConfig string. Then the main build function could be a match statement:
```gdscript
match game_config_version:
   GameConfigVersion.Latest:
       get_game_config_v8_text()
   GameConfigVersion.Version4:
       get_game_config_v4_text()
   GameConfigVersion.Version8:
       get_game_config_v8_text()
   _:
       printerr("Unsupported Game Config Version!")
```

I may see about adding it in myself after all, since you got the Version 4 text here and it doesn't seem like it'd be too much effort on my end.

Pesquisador-Stefano:
> I would avoid using text files or multiple files outside of the class. Instead I'd opt to keep everything contained within the TrenchbroomGameConfig resource script.
> 
> The way it's set up seems overdesigned, seems like it could break configs with team projects, and it's also using the TrenchBroom version numbers. Instead it should use the GameConfig Version numbers, as they don't always change when the editor version does.
> 
> Version should be determined with a `var game_config_version: int = 0`. I'd also recommend an enum instead to determine the GameConfig version numbers, in ascending order. Something like:
> 
> ```gdscript
> enum GameConfigVersion {
>     Latest,
>     Version4,
>     Version8
> }
> ```
> 
> Then the user would select the `game_config_version` in a dropdown.
> 
> Each version should have its own dedicated function that returns the GameConfig string. Then the main build function could be a match statement:
> 
> ```gdscript
> match game_config_version:
>    GameConfigVersion.Latest:
>        get_game_config_v8_text()
>    GameConfigVersion.Version4:
>        get_game_config_v4_text()
>    GameConfigVersion.Version8:
>        get_game_config_v8_text()
>    _:
>        printerr("Unsupported Game Config Version!")
> ```
> 
> I may see about adding it in myself after all, since you got the Version 4 text here and it doesn't seem like it'd be too much effort on my end.

done i think

addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
```diff
@@ -43,14 +50,21 @@ extends Resource
 @export_category("Editor Hint Tags")
 
 ## TrenchBroomTag resources that apply to brush entities.
-@export var brush_tags : Array[Resource] = []
+@export var brush_tags : Array[Resource] = [
+	preload("res://addons/func_godot/game_config/trenchbroom/tb_brush_tag_func.tres"),
+	preload("res://addons/func_godot/game_config/trenchbroom/tb_brush_tag_trigger.tres")
+]
```
RhapsodyInGeek, line 56: Brush tags should be empty, no default brush tags. We don't assume a user will be using `func_` or `trigger_` prefixes for their entities.

addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
```diff
@@ -43,14 +50,21 @@ extends Resource
...
 
 ## TrenchBroomTag resources that apply to brush faces.
 @export var brushface_tags : Array[Resource] = [
 	preload("res://addons/func_godot/game_config/trenchbroom/tb_face_tag_clip.tres"),
 	preload("res://addons/func_godot/game_config/trenchbroom/tb_face_tag_skip.tres")
 ]
 
+@export_category("Compatibility")
+## Keeps track of current Trenchbroom version. It's located above the New Map button when you open Trenchbroom.
```
RhapsodyInGeek, line 65: That's not an accurate description anymore.
`## Game configuration format compatible with the version of TrenchBroom being used.` would be more accurate. 
Avoid tutorialization in the property descriptions.

addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
```diff
@@ -86,44 +100,7 @@ func build_class_text() -> String:
...
-	"faceattribs": { 
-		"defaults": {
-			%s
-		},
-		"contentflags": [],
-		"surfaceflags": []
-	}
-}
-"""
+	var config_text : String = get_compatible_game_config_text()
```
RhapsodyInGeek, line 103: `get_compatible_game_config_text()` is superfluous indirection. Instead this could be where the match statement I recommended earlier would go:
```gdscript
	var config_text : String = ""
	match game_config_version:
		GameConfigVersion.Latest:
			config_text = get_game_config_v8_text()
		GameConfigVersion.Version4:
			config_text = get_game_config_v4_text()
		GameConfigVersion.Version8:
			config_text = get_game_config_v8_text()
		_:
			push_error("Unsupported Game Config Version!")

RhapsodyInGeek:
Sorry for the delay, but the PR looks good! Thanks for the contribution!

PR #32, opened 2024-09-02
Added Nearest/Linear Filtering Option for Materials Generated

skyelynwaddell:
I updated a feature to select which type of sampling you would like the materials to be generated as.
This is useful because Quake mappers may enjoy to use Nearest, and users familiar with Gold Source may enjoy to use Linear more.

Updates:
//func_godot_map_settings
```
## If true, all materials will use Nearest sampling, if false materials will use Linear
@export var nearest_filtering: bool = true
```

```
//func_godot_texture_loader
line 100:
#Set Material Texture Sampling Filter
		if map_settings.nearest_filtering == true:
			loaded_material.set_texture_filter(BaseMaterial3D.TEXTURE_FILTER_NEAREST)
		else:
			loaded_material.set_texture_filter(BaseMaterial3D.TEXTURE_FILTER_LINEAR)

```

### DIFF
--- a/addons/func_godot/src/map/func_godot_map_settings.gd
+++ b/addons/func_godot/src/map/func_godot_map_settings.gd
@@ -62,6 +62,9 @@ extends Resource
 ## If true, all materials will be unshaded, ignoring light. Also known as "fullbright".
 @export var unshaded: bool = false
 
+## If true, all materials will use Nearest sampling, if false materials will use Linear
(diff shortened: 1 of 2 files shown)


RhapsodyInGeek:
I see that you've already closed this, but I do want to say something like this won't be merged or implemented. All material settings should be set up in your materials, not the map settings, including the Default Material assigned in Map Settings.

PR #38, opened 2024-09-08
Adds Half-Life style 'Origin' texture support to set entity origin via a brush

jpiolho:
This PR adds [Half-Life style 'origin' texture support](https://twhl.info/wiki/page/ORIGIN) to easily set the origin point of an entity via a special texture brush. Adds a new entity origin type `Brush` to be able to set this

image

https://github.com/user-attachments/assets/fd005d10-864c-446d-af45-2be7be3461fb

---
# How to use

1. Make your entity
2. Create a brush where all the faces are textured with the special origin texture
3. Make this brush a part of the entity
4. In your func_godot FGD class definition, set the origin type to `Brush`
5. When you Build the map, the entity origin will be in the center of the geometry of the brush with the special texture

---

I was attempting to do a Gridmap tileset with different entities and found the alignment difficult to achieve. It's possible using Relative origin type, however I found it quite slow to figure out the correct value and the center of the entity didn't always match with what I was expecting, not to mention I'd have to update this values everytime I would update the model. So placing an origin brush avoids any ambiguity of where the origin should be. The classical use-case for this is for rotating doors style entities as shown above.

I'm unfamiliar with the codebase and gdscript so this implementation might not be the best, but it works. Also, I wasn't able to find the original font for the special textures, so the origin texture might look a bit off compared to the other ones


RhapsodyInGeek:
Took a look over this and some very cool stuff. I still have to scrutinize this obnoxiously before I merge it, especially since it modifies a lot of interconnected components, but I appreciate that you seem to have followed the design patterns throughout the plugin. Between adding the Brush OriginType to adding the Origin Texture to Map Settings alongside the Clip and Skip assignments, the design seems good. Integrated into existing systems, not trying to hackaround.

As far as the unpredictability of the default origin assignment, I agree. One change I've been meaning to implement that's really a carry over from Qodot is the way origin is determined. Shifty, the original designer for Qodot that FuncGodot is a descendant of, designed it to use the averaged center of all vertices. This makes non-cuboid origins difficult to determine prior to building.

I plan on adding an AABB origin determination at some point in the near future, to alleviate this unpredictability. I don't know whether anybody currently relies on the unpredictability of origin determination in their scripts, but I'd rather get rid of the averaged vertices method. What I'll likely do though is add AABB Center as an option, and then set that as the default.

I can't give this an in-depth look right now, but I'll put it on my priorities. Very cool addition that had been talked about before. Thanks for the contribution!

sinewavey:
I took a look this morning as well, and admittedly with a "isn't this already a thing?" approach - but it looked slotted in right to me as well and I like this especially to give an entry point into origin editing that aligns with well known mapping standards (Origin brushes for doors or platforms, etc). 

It's worth mentioning as this touches geo_gen @RhapsodyInGeek - perhaps something to consider for improvements there as have been previously desired. I think maybe some work can be done up front with things like this rather than inside that loop, but ultimately, that's a way larger change for something we've only briefly discussed and I think far exceeds the scope of this PR.

As a personal note: I've always felt using origin brushes rather than points to define a hinge or rotational axis was a bit odd.
When I implemented a similar feature for my own codebase, I found a point entity with an `angles` key to be really nice to define a new `info_axis` that was much more aligned with Hinge and Joints in Godot - and could be used directly as the origin rather than constructing a brush or AABB or any further calculations than reading the Vector3 position/rotation values directly.

All the above in mind, I like this addition and believe in the addition of it and wouldn't say no.

RhapsodyInGeek:
The problem I have with `info_axis` is that I don't see a way to associate the `info_axis` instance with a brush entity the same way the origin brush does this during the build process, without any class properties like `targetname` or some such thing. The main commonality of all of the default brush entities is that, other than the `Phong` base class, they don't have any associated scripts, they're just build variations. They also have no dependencies on other brushes, and the only design pattern they occupy are non-standard `_phong` and `_phong_angle` settings that almost no folks are using in their regular class properties.

Worth a discussion thread all on its own, don't want to clutter this PR talking about it too much.

What I want to say though is I like that the steps involved in setting an origin brush are simple and self-contained, as part of the natural build process as a Skip or Clip texture.

Question for @jpiolho : what's the fallback if a Brush Origin entity doesn't have an Origin brush? I'm sure I'd find out if I looked further, just don't really have the time or focus to study it at this moment.

sinewavey:
Absolutely. I don't mean to suggest my own solution, either - just that this subject is another thing to think about when we get to the eventual task. 


Wouldn't the fallback be nothing additional happening to entity.center? It looks like otherwise it just uses the `origin` property. In the changes to `geo_generator` this just adds on top of what we already do, assuming it did not get outright rejected from faces not matching (section above in generate_vertices_task)

jpiolho:
@RhapsodyInGeek 
> what's the fallback if a Brush Origin entity doesn't have an Origin brush?

It should be the same as origin type 'Ignore'. There's a condition [here](https://github.com/func-godot/func_godot_plugin/pull/38/files#diff-998cddf46d1450593c5377ddbbbdcc43ffacd22a80988759f621034eabc5a994R282) thats works because the `origin_mins` don't get modified unless there's an origin brush

addons/func_godot/src/map/func_godot_map_settings.gd
```diff
@@ -28,6 +28,9 @@ extends Resource
 ## Optional path for the skip texture, relative to [member base_texture_dir]. Brush faces textured with the skip texture will have those Faces removed from the generated [MeshInstance3D]. If the [FuncGodotFGDSolidClass] `collision_shape_type` is set to concave then it will also remove collision from those faces in the generated [CollisionShape3D].
 @export var skip_texture: String = "special/skip"
 
+## Optional path for the origin texture, relative to [member base_texture_dir]. Brush faces textured with the clip texture will have those Faces removed from the generated [MeshInstance3D]. The bounds of these faces will be used to calculate the origin point of the entity.
```
RhapsodyInGeek, line 31: Should read "Brush faces textured with the origin texture"

RhapsodyInGeek:
Just tested it out. Other than that minor comment mistake it all tests good and fits with the established styles for both the code and the texture. Good job and thanks for making the plugin better!

Issue #39, opened 2024-09-10
disabled entities on TB layers still invoked on map build


azur-wolve:
if a `tb_layer` that is disabled and marked as hidden (eye icon) from TB (using TB groups hierarchy)
FuncGodot still calls its `_func_godot_apply_properties()` and `_func_godot_build_complete()`

discovered when instancing nodes during map build in the editor from fgd generated nodes that were on disabled layers
Those dont appear on the scene dock but still on background seems to be

expected behavior to be disabled

RhapsodyInGeek:
Resolved by #45

PR #45, opened 2024-10-10
Rework TrenchBroom Layer Omission to Not Generate Entities

### LINKED ISSUE #31: let TB layers marked as disabled discard their contents on map build, even if TB groups hierarchy disabled
**_feature request_**

Screenshot from 2024-08-31 06-06-15

that would allow discard auxiliary layers used for prefabs, references, placeholder maps, etc

independently that if `FuncGodotMapSettings.use_trenchbroom_groups_hierarchy` is true or false

RhapsodyInGeek:
Old method used to generate nodes and free them at the end of the build. With this PR, FuncGodot now skips generating those nodes entirely.

It also contains some tweaks for performance optimization:
- Setting the size of the `entity_nodes` array to the `entity_dicts` size instead of appending in the entity loop.
- Instead of starting with a Node3D and then freeing it if the Node class needs to change according to the Entity Definition, it now only generates a Node3D if the Entity Definition *doesn't* generate a node.

Should resolve #31 and #39.

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -147,6 +147,9 @@ func run_build_step(step_name: String) -> Variant:
 
 ## Add [code]node[/code] as a child of parent, or as a child of [code]below[/code] if non-null. Also queue for ownership assignment.
 func add_child_editor(parent: Node, node: Node, below: Node = null) -> void:
+	if not node or not parent:
+		return
+	
 	var prev_parent = node.get_parent()
 	if prev_parent:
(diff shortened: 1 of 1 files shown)


sinewavey (approved):
apologies for the delay. looks good - it strikes me as a touch odd leaving null entries in `entity_nodes` but it makes sense and there isn't a real reason to recompute it so much

RhapsodyInGeek:
It seems odd, but it's because the `entity_nodes` array needs to align with the `entity_dicts` array, which is one of those "house of cards" situations.

PR #56, opened 2024-11-05
Allows automatically applying Node properties directly to the Node. 

greenfox1505:
Adds a override_node_properties bool to FGD Entity. Map generator read apply this value directly to a node if there is a matching property name. If that property is a different type, it will cause an map build error.

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -22,6 +22,9 @@ var prefix: String = ""
 ## Key value pair properties that will appear in the map editor. After building the FuncGodotMap in Godot, these properties will be added to a Dictionary that gets applied to the generated Node, as long as that Node is a tool script with an exported `func_godot_properties` Dictionary.
 @export var class_properties : Dictionary = {}
 
+## Used to automatically apply properties to node.
+@export var override_node_properties : bool = false
+
 ## Descriptions for previously defined key value pair properties.
 @export var class_property_descriptions : Dictionary = {}
 
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -22,11 +22,11 @@ var prefix: String = ""
 ## Key value pair properties that will appear in the map editor. After building the FuncGodotMap in Godot, these properties will be added to a Dictionary that gets applied to the generated Node, as long as that Node is a tool script with an exported `func_godot_properties` Dictionary.
 @export var class_properties : Dictionary = {}
 
-## Used to automatically apply properties to node.
-@export var override_node_properties : bool = false
-
 ## Descriptions for previously defined key value pair properties.
 @export var class_property_descriptions : Dictionary = {}
+
+## Used to automatically apply properties to node.
+@export var auto_apply_to_matching_node_properties : bool = false
 
 ## Appearance properties for the map editor. See the [**Valve FGD**](https://developer.valvesoftware.com/wiki/FGD#Entity_Description) and [**TrenchBroom**](https://trenchbroom.github.io/manual/latest/#display-models-for-entities) documentation for more information.
 @export var meta_properties : Dictionary = {
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -25,7 +25,7 @@ var prefix: String = ""
 ## Descriptions for previously defined key value pair properties.
 @export var class_property_descriptions : Dictionary = {}
 
-## Used to automatically apply properties to node.
+## Automatically applies entity class properties to matching properties in the generated node. When using this feature, class properties need to be the correct type or you may run into errors on map build.
 @export var auto_apply_to_matching_node_properties : bool = false
 
 ## Appearance properties for the map editor. See the [**Valve FGD**](https://developer.valvesoftware.com/wiki/FGD#Entity_Description) and [**TrenchBroom**](https://trenchbroom.github.io/manual/latest/#display-models-for-entities) documentation for more information.
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -1007,6 +1007,11 @@ func apply_properties_and_finish() -> void:
 		
 		if 'func_godot_properties' in entity_node:
 			entity_node.func_godot_properties = properties
+
+		if override_node_properties:
+			for i in properties:
+				if i in entity_node:
+					entity_node[i] = properties[i]
 		
 		if entity_node.has_method("_func_godot_apply_properties"):
 			entity_node.call("_func_godot_apply_properties", properties)
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -1008,7 +1008,7 @@ func apply_properties_and_finish() -> void:
 		if 'func_godot_properties' in entity_node:
 			entity_node.func_godot_properties = properties
 
-		if override_node_properties:
+		if auto_apply_to_matching_node_properties:
 			for i in properties:
 				if i in entity_node:
 					entity_node[i] = properties[i]
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -1009,9 +1009,12 @@ func apply_properties_and_finish() -> void:
 			entity_node.func_godot_properties = properties
 
 		if auto_apply_to_matching_node_properties:
-			for i in properties:
-				if i in entity_node:
-					entity_node[i] = properties[i]
+			for property in properties:
+				if property in entity_node:
+					if typeof(entity_node.get(property)) == typeof(properties[property]):
+						entity_node.set(property, properties[property])
+					else:
+						push_error("Entity %s property \'%s\' type mismatch with matching generated node property." % [entity_node.name, property])
 		
 		if entity_node.has_method("_func_godot_apply_properties"):
 			entity_node.call("_func_godot_apply_properties", properties)


addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
```diff
@@ -22,6 +22,9 @@ var prefix: String = ""
 ## Key value pair properties that will appear in the map editor. After building the FuncGodotMap in Godot, these properties will be added to a Dictionary that gets applied to the generated Node, as long as that Node is a tool script with an exported `func_godot_properties` Dictionary.
 @export var class_properties : Dictionary = {}
 
+## Used to automatically apply properties to node.
+@export var override_node_properties : bool = false
```
RhapsodyInGeek, line 26: Should be ordered at the bottom, below descriptions, due to descriptions being both more important and having a more direct relationship to the entity definition, in addition to being "the next step" in constructing class properties.

Also, I think we should call this `auto_apply_to_matching_node_properties`, as it's both more descriptive and more accurate to what's happening. I'm not much a fan of the term "override" in this particular case, since we're setting the properties for the first time on the newly generated node.

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -1007,6 +1007,11 @@ func apply_properties_and_finish() -> void:
 		
 		if 'func_godot_properties' in entity_node:
 			entity_node.func_godot_properties = properties
+
+		if override_node_properties:
+			for i in properties:
```
RhapsodyInGeek, line 1012: Should replace `i` with `property`, to match earlier conventions and for clarity.

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -1007,6 +1007,11 @@ func apply_properties_and_finish() -> void:
...
+				if i in entity_node:
+					entity_node[i] = properties[i]
```
RhapsodyInGeek, line 1014: Does `node[property] = x` run the node's setter method for that property? Does this work for custom class properties?

Prefer `entity_node.set(properties[property])`, unless there's a significant performance improvement to the current implementation (as long as it also still runs the setter method).

greenfox1505: It does work for class properties. I'm using this for all my lights, setting brightness, color, shadows on/off, etc. No need for any GDScripts in the FGD to directly control built in Godot Nodes. It'll run GDScript set/gets too.

To use `Object.set(` you'll need is:

```
entity_node.set(property, properties[property])
```

I don't usually use `object.set(`; I prefer the `[]` syntax. In general, I feel like it reads easier. I don't know if there is a performance implication, but it's certainly testable. I'd imagine it'd tokenize faster, but probably has basically the same runtime; I speculate (based only on how I would do it) it parses as the same underlying call. 

It doesn't check to see if it's the right type, but it will fail if it's the wrong type. Unless it's no type. `var a` will always succeed, but `var b:Vector3` `will only succeed if it'd a Vector3 (maybe if it's castable to a Vector3 it works?)

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
greenfox1505: Well, it is overriding the default prefab or class. But I'm not opposed to changing the name or moving it.

addons/func_godot/src/map/func_godot_map.gd
RhapsodyInGeek: > It does work for class properties. I'm using this for all my lights, setting brightness, color, shadows on/off, etc. No need for any GDScripts in the FGD to directly control built in Godot Nodes. It'll run GDScript set/gets too.
I wasn't asking about the built in Node class properties, I was asking about custom classes, like say I have a `LightDx` with an `@export var style` that uses a setter function to modify some other private variable somewhere.

> To use Object.set( you'll need is: `entity_node.set(property, properties[property])`
Ah yeah, my bad, I was checking this out at around 2 or 3 in the morning.

>  I speculate (based only on how I would do it) it parses as the same underlying call.
Can you check and verify this? I prefer not to speculate.

In any case, we don't actually use `node[property] = value` anywhere else, but we do use `node.set(property, value)` elsewhere in FuncGodot. That's the GDScript standard way of setting properties by String, so we should stick with that convention regardless of individual preferences.

> It doesn't check to see if it's the right type, but it will fail if it's the wrong type. Unless it's no type. var a will always succeed, but var b:Vector3 `will only succeed if it'd a Vector3 (maybe if it's castable to a Vector3 it works?)
When I was thinking of type checking, I meant using the Variant class' type checking in general, not to double check for specific types but that the node property and the entity property are the same type. Can't remember the syntax off the top of my head but something like
```gdscript
typeof(node[property]) == typeof(properties[property])
```
We can then put an error message describing the problem as an else. Something like:
```gdscript
		if auto_apply_to_matching_node_properties:
			for property in properties:
				if property in entity_node:
					if typeof(entity_node.get(property)) == typeof(properties[property]):
						entity_node.set(property, properties[property])
					else:
						printerr("Entity %s property \'%s\' type mismatch with matching generated node property." % [entity_node.name, property])
```
That said, I can see where it might not work if it just considers all Dictionary properties to be of type Variant rather than the specific Variant inheriting class, since that's what Dictionaries and Arrays use for their value type unless specified.

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
greenfox1505: I'm putting `auto_apply_to_matching_node_properties` between meta properties and descriptions. IMHO, it should be as close to `class_properties` as possible since that is what it effects.

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
```diff
@@ -25,6 +25,9 @@ var prefix: String = ""
 ## Descriptions for previously defined key value pair properties.
 @export var class_property_descriptions : Dictionary = {}
 
+## Used to automatically apply properties to node.
```
RhapsodyInGeek, line 28: I think my previous comment on this description got lost. Should be changed to
```gdscript
## Automatically applies entity class properties to matching properties in the generated node. When using this feature, class properties need to be the correct type or you may run into errors on map build.
```

addons/func_godot/src/map/func_godot_map.gd
greenfox1505: > In any case, we don't actually use node[property] = value anywhere else, but we do use node.set(property, value) elsewhere in FuncGodot.

I only found this line: `addons/func_godot/src/util/func_godot_texture_loader.gd:139`:`material.set(enable_prop, true)`.

The reset of the project uses `[]` syntax for Dictionaries. Treating objects like Dictionaries and Dictionaries like Objects is a very Godot thing to do. I would substitute some of those `dict["string"] = etc` with `dict.string = etc`. But I'm not arguing for that. 

```gdscript
var itter = 5_000_000
var runs = 5

var setTest:="Empty":
	set(v):
		setTest = v
var noSetter:="Empty"

func _ready() -> void:
	failCheck()
	for i in runs:
		print("ITTERATION:",i)
		bench()
		print("=============")

func failCheck():
	#this silently fails. the value does not change by the debugger doesn't notice
	set("setTest",42)
	#this will fails hard
	self["setTest"] = 42
	#this won't run:
	setTest = 42


func bench() -> void:
	var t

	t = Time.get_ticks_usec()
	for i in itter:
		noSetter = "1234"
	print('noSetter = "notEmpty"      ',(Time.get_ticks_usec()-t)/1_000_000.0)


	t = Time.get_ticks_usec()
	for i in itter:
		setTest = "qwer"
	print('setTest = "qwer"           ',(Time.get_ticks_usec()-t)/1_000_000.0)

	t = Time.get_ticks_usec()
	for i in itter:
		self.setTest = "zxcv"
	print('self.setTest = "zxcv"      ',(Time.get_ticks_usec()-t)/1_000_000.0)

	t = Time.get_ticks_usec()
	for i in itter:
		self["setTest"] = "asdf"
	print('self["setTest"] = "asdf"   ',(Time.get_ticks_usec()-t)/1_000_000.0)

	t = Time.get_ticks_usec()
	for i in itter:
		set("setTest","qwer")
	print('set("setTest","qwer")      ',(Time.get_ticks_usec()-t)/1_000_000.0)

	t = Time.get_ticks_usec()
	for i in itter:
		self.set("setTest","type")
	print('self.set("setTest","type") ',(Time.get_ticks_usec()-t)/1_000_000.0)

	t = Time.get_ticks_usec()
	for i in itter:
		if(typeof("set(") == typeof("type")):
			set("setTest","type")
	print("type checking:             ",(Time.get_ticks_usec()-t)/1_000_000.0)

	setTest = "good"
	self.setTest = "bye"
	var dict = {}
	dict["world"] = "hello" #init 

	t = Time.get_ticks_usec()
	for i in itter:
		dict["world"] = "hello__"
	print('dict["world"] dt=          ',(Time.get_ticks_usec()-t)/1_000_000.0)

	t = Time.get_ticks_usec()
	for i in itter:
		dict.world = "goodbye"
	print("dict.world dt=             ",(Time.get_ticks_usec()-t)/1_000_000.0)
```

Went into a rabbit hole of benchmarking `.set` vs `[]`. Found this interesting quark. `[]` will trigger an error when setting the wrong type, `.set(` does nothing on a type missmatch. No error. No value change. I'm curious what it does in a release build, but not enough at the moment to check.

The performance between `.set(` and `[]` does indeed lean in the `set`'s favor, but only by a TINY margin (I haven't testing in a release build, which likely would effect results). Perhaps due to the error checking seen above. I haven't testing String vs StringName which also could make a difference.

But if performance is that high of a priority, we probably should skip the type checking. I would gently suggest just using `assert(` for type checking:

```gdscript
assert(typeof(entity_node.get(property)) == typeof(properties[property]),"Entity %s property \'%s\' type mismatch with matching generated node property." % [entity_node.name, property])
```

A debug build will fail reading a bad map, but it'll halt to tell you why it failed. A release build will just skip this check. (yay performance!) 

There are other places in FuncGodot where bad data cause maps to fail semi-quietly, and we could catch those with assertion checks. IMHO, Errors in FGD data should be considers hard stops at runtime.

RhapsodyInGeek: > I only found this line: addons/func_godot/src/util/func_godot_texture_loader.gd:139:material.set(enable_prop, true).

There are I think 3 instances of `set(property, value)`, one of which is actually in func_godot_map.gd. I believe the other 2 are in the texture loader files. There might be more, but those are the only ones that I found in a quick search earlier before I made that assertion about them. In any case, that's enough to establish a pattern and it follows suit with GDScript best practices.

> The reset of the project uses [] syntax for Dictionaries. Treating objects like Dictionaries and Dictionaries like Objects is a very Godot thing to do.

We use `dict[prop]` for Dictionaries because we don't always know the contents of a Dictionary, and GDScript standards tend towards `dict[prop]` over `dict.prop` or `dict.get(prop, default_value)`.

> [] will trigger an error when setting the wrong type, .set( does nothing on a type missmatch. No error. No value change. I'm curious what it does in a release build, but not enough at the moment to check.

Well that kinda clinches it even more. Definitely using the `set` method.

> There are other places in FuncGodot where bad data cause maps to fail semi-quietly, and we could catch those with assertion checks. IMHO, Errors in FGD data should be considers hard stops at runtime.

Build process is a bit of a house of cards if I'm being honest. Need to add a button to reset the build environment in case the build process breaks and softlocks the FuncGodotMap, but that's best saved for another PR.

I disagree that errors should be hard stops. They should be reported in the output log, but it can be important to see what all still works even if you make a mistake somewhere. It doesn't actually hurt to let the map building finish as long as we're telling people what's going wrong. The preference is to catch bad data and replace it with default values while logging the error, not to hard stop.

In this particular case while it's nice to know that with `set()` we don't need to worry, I would still rather we go with an `if else` type check, not an `assert`. Performance _is_ important, but I think a Variant type check should be fine here since it's a simple comparison and it's on a different build step than where I'm really struggling for performance (GeoGenerator).

greenfox1505: This is not just bad data coming for a bad map file or Trnechbroom in some way. The only way for it to get to this line and there to be a type error is if your FGD Entity properties and your Godot class don't match type. If you've missed matched types that's pretty serous internal configuration error, not just bad map data. IMHO, this is a big enough problem to halt for, either by assert or type conflict. Every other place where you have a type conflict in the engine, its a halting error (either failure to parse/run or halt at runtime). A `printerr(` does not match the severity of this type of data conflict.

I can implement this how you want. But I do strongly believe a type error in the devs FGD data entry is worth halting over.

RhapsodyInGeek: > Every other place where you have a type conflict in the engine, its a halting error (either failure to parse/run or halt at runtime).

This isn't really true though. You even said it yourself: the `set` method has a built in mechanism to avoid setting incorrect values.

This is how Godot handles the `set` method:

https://github.com/godotengine/godot/blob/87318a2fb7fffeb72adca934e31915be077c3d1f/core/object/object.cpp#L228

https://github.com/godotengine/godot/blob/master/core/object/class_db.cpp#L1582

> A printerr( does not match the severity of this type of data conflict. I can implement this how you want. But I do strongly believe a type error in the devs FGD data entry is worth halting over.

Please implement it without the assert, using an `if else` design that allows it to finish applying properties. It's not necessary to halt the build process in this case (and applying properties is part of the build process) especially since they will still have their default values after failing.

On a related note, recently I watched a game dev conference talk from John Romero, and one of his big tips was (paraphrased) "_Don't let the game crash_; if data fails to load, make sure to provide a default value, especially a value that will let you know the correct data failed to load."

That's partly why instead of halting the build process when FuncGodot can't find a texture or material, it instead uses the Green Checkerboard. It's also partly why if properties are missing from the Map file due to not being filled out, FuncGodot automatically adds them and supplies default values to the properties dictionary.

There isn't sufficient reason in this particular case to halt the map build over just letting it run through and pointing out FGD errors. We should only halt the map build if it were to crash the engine.

greenfox1505:
I hope it's okay that I used `push_error(` instead. This is a louder error then `printerr(` and will help devs find those errors without halting.

image

RhapsodyInGeek:
While I would like errors to be reported uniformly, it may be worth changing all printerrs over to push_error. We'll have to make a decision on that in the near future. For now, push_error works fine.

RhapsodyInGeek:
Looks good!

PR #57, opened 2024-11-05
Typed `Array[Resource]` to Arrays of typed classes

greenfox1505:
Changes `Array[Resource]` to expected types. Makes it easier to add new assets. But blocks dragged existing assets in.

This bug engine bug is blocking this merge from being accepted: https://github.com/godotengine/godot/issues/92666


RhapsodyInGeek:
This PR is no longer compatible with the current version of FuncGodot. That said, as soon as Godot resolves the drag 'n drop custom resources issue the team will implement static typing to all applicable arrays where possible.

PR #78, opened 2025-01-21
Update TrenchbroomGameConfig Latest Version Default to Version 9

greenfox1505:
Updated default GameConfig to Version9, since Trenchbroom 2024.2 has been stable for 3 weeks.

### DIFF
--- a/addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
+++ b/addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
@@ -170,7 +170,7 @@ func parse_tags(tags: Array) -> String:
 		tags_str += "\t\t\t}"
 		if brush_tag != tags[-1]:
 			tags_str += ","
-	if game_config_version < GameConfigVersion.Version9:
+	if game_config_version == GameConfigVersion.Version8:
 		tags_str = tags_str.replace("material", "texture")
 	return tags_str
 
--- a/addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
+++ b/addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
@@ -322,7 +322,7 @@ func get_game_config_v9v8_text() -> String:
 }
 	"""
 	
-	if game_config_version != GameConfigVersion.Version9: # Change this to check if == Version8 when TB 2024.2 hits Stable
+	if game_config_version == GameConfigVersion.Version8:
 		config_text = config_text.replace(": 9,", ": 8,")
 		config_text = config_text.replace("material", "texture")
 	


addons/func_godot/src/trenchbroom/trenchbroom_game_config.gd
```diff
@@ -170,7 +170,7 @@ func parse_tags(tags: Array) -> String:
 		tags_str += "\t\t\t}"
 		if brush_tag != tags[-1]:
 			tags_str += ","
-	if game_config_version < GameConfigVersion.Version9:
+	if game_config_version == GameConfigVersion.Version8:
```
RhapsodyInGeek, line 173: This will break Version 4 compatibility and needs to be reverted back to what it was previously.

greenfox1505: Maybe != Version9? It needs to say "material" on v9

greenfox1505: I did that because it broke on my project.

greenfox1505: Latest and 9 should say "material".

RhapsodyInGeek: > Maybe != Version9?

Will break future version support which will use the "material" tag.

IIRC it's because the TrenchBroomTag resources use "material" by default since Version 9 support got added, and need to be switched back to "texture" for previous versions.

RhapsodyInGeek: I think I know why it breaks, actually. It's because the enum for Latest is 0.

So it should be changed to:
`if game_config_version != GameConfigVersion.Latest and game_config_version < GameConfigVersion.Version9:`

greenfox1505: Yeah, I was gonna say that, but I'm at dinner.

We if you want to modify those values, I think that enum should be explicate values. Latest = 0, etc. I'd suggest v4=4 too, but that'll break existing serialization 

I'd probably do "== v4 or == v8".

RhapsodyInGeek: That works until someone decides for some reason they want to add support for the other versions for some reason or another. Also, not that it actually matters in this specific case (though it might in other areas of the plugin), but `x < z` is more performant than `x == y`.

Enum values can't be modified or else we risk breaking some folks' projects for no actual good reason, especially when it's easily avoided. New versions needed to be appended to the end, we won't be inserting them at the front. 

`if game_config_version != GameConfigVersion.Latest and game_config_version < GameConfigVersion.Version9:` is what it's gotta be. It works, it's foreseeable future proofed, and it only breaks compat on people still using the "Latest" setting if they're not prepared for it.

Fun bit of FuncGodot trivia: I was just telling Xage the whole reason the compatibility mode exists is because someone needed FuncGodot to work with Game Config Version 4, and we ported over the config text from an older version of Qodot to add it in (FuncGodot launched with Version 8 support and Qodot's last version dropped Version 4 support for 8).

RhapsodyInGeek:
Implemented by #80

PR #86, opened 2025-02-23
Scale Model Point Entities when Scale Expression set

hryx:
The _scale expression_ setting can be used to allow a scaling key to be applied to entities in TrenchBroom, but the map built by func_godot does not reflect the size as shown in the map editor.

This change scales the generated node if a scale expression is defined and the entity's properties include a float value for that key. The final scale is affected by the map's inverse scale factor.

For example, with the default map inverse scale factor of 32 and a scaling expression of `scale`, an entity tagged with `scale = 8` in TrenchBroom results in a node scaled by one quarter.

shot 2025-02-23 15 30 19@2x

This is an unsolicited change so please do let me know if this is misguided or won't work for some reason.

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -455,6 +455,19 @@ func build_entity_nodes() -> Array:
 							angles.y += angle
 						angles.y += 180
 						node.rotation_degrees = angles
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
Feature is now supported via 4b6d60d

PR #87, opened 2025-02-26
Fix invalid container id when unregistering progress bar control

pminiszewski:
This PR fixes an error message being displayed when closing Godot editor or disabling func_godot plugin in the project settings: 
` scene/main/node.cpp:1602 - Condition "p_child->data.parent != this" is true.`


RhapsodyInGeek:
Looks good!

PR #88, opened 2025-03-11
Add support for custom classes

sol-vin:
Adds support for providing custom classes to entities via Node Generation -> Node Class, `ClassDB` doesn't register custom classes, but using this method we can ensure they are found.

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -352,6 +352,18 @@ func build_texture_size_dict() -> Dictionary:
 			texture_size_dict[tex_key] = Vector2.ONE
 	
 	return texture_size_dict
+	
+static func get_script_by_name(name_of_class:String) -> Script:
+	if ResourceLoader.exists(name_of_class, "Script"):
+		return load(name_of_class) as Script
+
+	for global_class in ProjectSettings.get_global_class_list():
+		var found_name_of_class:String = global_class["class"]
+		var found_path:String = global_class["path"]
+		if found_name_of_class == name_of_class:
+			return load(found_path) as Script
+
+	return null
 
 ## Build nodes from the entities in [member entity_dicts]
 func build_entity_nodes() -> Array:
@@ -423,7 +435,12 @@ func build_entity_nodes() -> Array:
 						entity_nodes[entity_idx] = null
 						continue
 					if entity_definition.node_class != "":
-						node = ClassDB.instantiate(entity_definition.node_class)
+						if ClassDB.class_exists(entity_definition.node_class):
+							node = ClassDB.instantiate(entity_definition.node_class)
+						else:
+							var script := get_script_by_name(entity_definition.node_class)
+							if script is GDScript:
+								node = (script as GDScript).new()
 				elif entity_definition is FuncGodotFGDPointClass:
 					if entity_definition.scene_file:
 						var flag: PackedScene.GenEditState = PackedScene.GEN_EDIT_STATE_DISABLED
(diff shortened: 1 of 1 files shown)


addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -352,6 +352,18 @@ func build_texture_size_dict() -> Dictionary:
 			texture_size_dict[tex_key] = Vector2.ONE
 	
 	return texture_size_dict
+	
+static func get_script_by_name(name_of_class:String) -> Script:
```
RhapsodyInGeek, line 356: Function should be called `get_script_by_class_name`

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -352,6 +352,18 @@ func build_texture_size_dict() -> Dictionary:
...
+	if ResourceLoader.exists(name_of_class, "Script"):
+		return load(name_of_class) as Script
+
+	for global_class in ProjectSettings.get_global_class_list():
+		var found_name_of_class:String = global_class["class"]
```
RhapsodyInGeek, line 361: May seem a little thing, but please make sure you put a space between the var name and type, i.e:
`var found_name_of_class: String = global_class["class"]`

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -423,15 +435,25 @@ func build_entity_nodes() -> Array:
 						entity_nodes[entity_idx] = null
 						continue
 					if entity_definition.node_class != "":
-						node = ClassDB.instantiate(entity_definition.node_class)
+						if ClassDB.class_exists(entity_definition.node_class):
+							node = ClassDB.instantiate(entity_definition.node_class)
+						else:
+							var script := get_script_by_name(entity_definition.node_class)
```
RhapsodyInGeek, line 441: Try to make sure variables are as explicit to what they are as possible.
Can this be `var script: Script = get_script_by_name(entity_definition.node_class)` instead or does that break the check for `GDScript` below?

sol-vin: Fine to change it but from my understanding `:=` does the same thing.

sol-vin:
Suggestions handled 🫡

azur-wolve:
then this allows to use custom class names on `FuncGodotFGDEntityClass.node_class` on the inspector

addons/func_godot/src/map/func_godot_map.gd
RhapsodyInGeek: Part of the reason is not just for strong static typing, but also for easier readability. While I can look up what that function returns, making sure that it explicitly says that it returns `Script` means other contributors don't have to go bouncing around in a plugin that already forces them to do just that for too many other things. It's also part of the code style we've adopted for FuncGodot, and we want to make sure contributions follow the same style, once again for easier readability.

RhapsodyInGeek:
> Suggestions handled 🫡

Cool. Gonna test it out tomorrow since it's late right now, but looking over it I'm sure it'll be a quick merge. Thanks for the contribution!

sol-vin:
Fat finger closed my bad :( 👎🏻

RhapsodyInGeek:
Sorry, still haven't tested this yet. Been busy. I'll see if I can get to it soon, unless @sinewavey might be able to test it? Looks like it _should_ work, but still want to verify.

sinewavey:
Hi, sorry for the delay.
I've been able to test this with success so far.

It also works with user classes as a base and a non-class named extension set in "script_class" as well; however, if the two do not match, the entity is forced to change type seemingly.

I don't think this is is strictly something needed to be addressed - similar things happen when trying to export an incorrect FGD, or without a path set.

Overall, I think this is a nice change.

RhapsodyInGeek:
Looks good to me.

PR #90, opened 2025-03-26
add unique_name bool to generate "scene unique" node names

GammaGames:
apologies for the trimmed whitespace

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -44,37 +44,39 @@ var prefix: String = ""
 ## Nodes will be named `"entity_" + name_property`. An entity's name should be unique, otherwise you may run into unexpected behavior.
 @export var name_property := ""
 
+@export var unique_name : bool = false
+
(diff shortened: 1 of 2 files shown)


RhapsodyInGeek:
When submitting a PR you need to explain what this PR is for, what it does, why it's needed. Simply linking to the official docs is not enough.

You also need to make sure your property is well commented so that users know exactly what it does. Try to follow the style and verbosity of other properties. Commenting shouldn't be necessary in the code if you write your code in the same style and manner as the rest of the FuncGodot code.

You'll also need to change the property name to something that succinctly describes what it actually does.

Lastly, this should also be added to the _FuncGodotMapSettings_ resource, which decides the default behavior for all generated entities. The property in the entity definition should be an override. See how the name property is handled as an example.

I suspect that you'll need to make the entity definition property an enum for it to act as an override to Map Settings. My recommendation would be:
```
enum SceneUniqueNameMode { INHERIT, FALSE, TRUE }
```
The Map Settings property would only need to be a bool though.

GammaGames:
These changes are untested, I did this during my lunch break and will try them out tonight

RhapsodyInGeek:
Please edit your original post to include what it's for, what it does, and what you did / are doing.

Don't currently have time now to test it and review further but I'll try to get to it soon.

GammaGames:
Hopefully better, thanks for the patience!

PR #99, opened 2025-05-01
allow_String-StringName_conversion_on_map_build

azur-wolve:
when using `FuncGodotFGDEntityClass.auto_apply_to_matching_node_properties`

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -1138,7 +1138,13 @@ func apply_properties_and_finish() -> void:
 				if entity_definition.auto_apply_to_matching_node_properties:
 					for property in properties:
 						if property in entity_node:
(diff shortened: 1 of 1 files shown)


sinewavey:
Of the two PRs regarding type conversion, I think this one's the most agreeable to start off with.

I'll have to look a little closer myself for further type conversions on other values

azur-wolve:
the other PR has more bulky code
but conversions from `String` to generated node var type come handy for properties created from TB for sure
just ask if have any questions

RhapsodyInGeek:
The core rewrite nullifies this PR. Additionally, StringNames are now accounted for as of the latest FuncGodot release. While they're still not considered during `auto_apply_matching_properties`, this behavior is consistent with other property types that are "similar but different", and as such we won't be automatically converting Strings to StringNames. Instead, users should make sure to set the correct variant type in the entity definition's class properties.

Issue #98, opened 2025-05-01
Bypass StringName properties problems & Allow Godot implicit conversions with FuncGodotFGDEntityClass.auto_apply_to_matching_node_properties


azur-wolve:
### Problem 1

It looks that TrenchBroom doesnt like `StringName` on the fgd properties via `FuncGodotFGDEntityClass.class_properties` dictionary, which is allowed by the godot editor exported dictionary UI while choosing types.
Such properties wont appear in TrenchBroom.

**Solution:**
That could be addressed by converting to `String` when generating the fgd definitions for TB.

### Problem 2

When have a custom property of type `String` applied to a node property of type `StringName` via `FuncGodotFGDEntityClass.auto_apply_to_matching_node_properties` it displays an error:
```
  ERROR: core/variant/variant_utility.cpp:1098 - Entity entity_X property 'WHATEVER' type mismatch with matching generated node property.
```

**Solution A:**
When auto applying properties via `auto_apply_to_matching_node_properties` a conversion from `String` to `StringName` if the property in the node is a `StringName`.

currently:
```gdscript
# func_godot.gd

# line 1116
if entity_definition.auto_apply_to_matching_node_properties:
	for property in properties:
		if property in entity_node:
			if typeof(entity_node.get(property)) == typeof(properties[property]): # HERE'S THE ISSUE
				entity_node.set(property, properties[property])
			else:
				push_error("Entity %s property \'%s\' type mismatch with matching generated node property." % [entity_node.name, property])
```

Not only will happen with `String` ←→ `StringName` but any expected implicit conversion in the way Godot does them, like between `int`, `float`, `bool`.

**Solution B:** (better)
if the name match AND the types have implicit conversion, do it.

PR #100, opened 2025-05-02
Add implicit Godot type conversions when using auto_apply_to_matchi…

azur-wolve:
solves https://github.com/func-godot/func_godot_plugin/issues/98 → **Problem 1, Solution B**

**IMPORTANT:** its done from the current 2025.1 downloaded from the asset store, sorry if its a bit of a mess.

This PR adds the function `try_handle_implicit_godot_type_conversion()` to `FuncGodotMap`.
It tries to handle implicit Godot type conversions gracefully.
Returns `bool` to indicate if the conversion was successful.
If not, will print the pre-existing error message of type mismatch.

Only called if `auto_apply_to_matching_node_properties` enabled.

**Note:** Conversions from `String` are handy as all properties created from TrenchBroom are strings.

---

handles the following conversions:
`[fgd_ent_property_type]` → `[generated_node_property_type]`

- `String` → `StringName` (so it supersedes https://github.com/func-godot/func_godot_plugin/issues/99)

- `String` → `bool`

- `String` → `int` (shows warning of truncation, only if there's decimal fraction, prints discarded fraction)

- `String` → `float`

- `String` → `Vector3`

- `String` → `Vector3i` (shows warning of truncation, only if there's any decimal fraction on any of the components, prints which components and the discarded fraction)

- `String` → `Vector2`

- `String` → `Vector2i`

- `String` → `Vector4`

- `String` → `Vector4i`

- `String` → `Color`

- `int` → `float`

- `bool` → `int`

- `int` → `bool`

- `bool` → `float`

- `float` → `bool`

- `float` → `int` (shows warning of truncation, only if there's decimal fraction, prints discarded fraction)

- `Vector2i` → `Vector2`

- `Vector3i` → `Vector3`

- `Vector4i` → `Vector4`

- `Vector3` → `Vector3i`

- `Vector2` → `Vector2i`

- `Vector4` → `Vector4i` 

- `Vector4` | `Vector4i` | `Vector3` | `Vector3i` → `Color` (could help to bypass the TB limitation of alpha channel being always 1.0, and the use of an additional float property for it)

**NOTE:** in all conversions to `Vector3i` / `Vector2i` / `Vector4i`, shows warning of truncation, only if there's any decimal fraction on any of the components, prints which components and the discarded fraction.

---

**missing conversions:**
    - `Dictionary` (choices)
    - `Array` (flags)
    - `NodePath` (target_destination)
    - `Material` (material / shader)
    - `Texture2D` (decal)
    - `AudioStream` (sound)
    - `Object` (target_source)
    - `bool` → String
    - `int` → String
    - `float` → String
 
 haven't done anything with those as didn't know what to do really
 if there would be some utility in converting them also just tell me and will add

---

### Vector conversions

all vec types work with the string formats:

    "n n n" / "n, n, n" / "n,n,n" / "(n n n)" / "(n, n, n)" / "(n,n,n)"

the only difference is the num of components

considered the notation: `x: {n} y: {n} z: {n} w: {n}` (even disordered or if nums are not separeted from ':')
but decided to drop those for simplicity and to dont overload the algorithm speed

**Note:**
i've used the Godot notation "X Y Z W" for vectors with 4 components
but in the FuncGodot docs appears as "W X Y Z"
if its incorrect just tell me and will change it

also as the fgd value types return strings for all vector types, have just commented the conversions where fgd property type corresponds to Godot built-in vectors, so are just operative the conversions `String` → `Vector*`
if those already have been converted on `entity_dicts` just tell and will un-comment them

---

### Color converisons

allowed formats:

    hexcode (HTML)

    rgb → "n n n" / "n, n, n" / "n,n,n" / "(n n n)" / "(n, n, n)" / "(n,n,n)"

    rgb + alpha → same but with an extra channel

    {colorname}
    
    {colorname} + alpha

  
  colorname are like Color constants (case insensitive)
  
  alpha in any case can be either rgb(0-255) or 0.0-1.0 (snapped to 0.001 like in Godot Editor inspector)
  color channels can be either rgb(0-255) or 0.0-1.0, the 3 must have the same format, if only one of them is > 1 all will be trated as rgb
  
  values will be clamped to those limits so negative numbers will become 0
  
  so can use any combination color + alpha like:

    255 255 255 1.0
    1.0 1.0 1.0 255
    red 255
    red 1.0
  
  hex color codes in Godot use RGBA so they have alpha included
  (what i dont know yet is if it works as expected if use hex codes without alpha channel encoded)

**Note:** some random words that are only composed by letters from a-f are considered colors, like "caca" which is shit in spanish, the color is a transparent purple instead of brown
  
  I thought in allowing other formats like:
      
     r: 255 g: 255 b: 255 a: 255
(even disordered and if nums not separate from ':')
      
  but decided to drop those for simplicity and to dont overload the algorithm speed

---

**Final notes:** 

I wasn't sure about `Vector4` | `Vector4i` → `Color` conversion, so it is commented for now. Decide if so.

there has been some effort into make the algorithm performant by distributing the conditional branches

also have tried to check the correctness of the values before apply them, but check just in case

not sure if the code is correct or not regarding the FuncGodot design philosophy

in case that the PR isnt ok, if there's something useful just grab and reuse it.

should choose between this and https://github.com/func-godot/func_godot_plugin/issues/99

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -1140,7 +1140,7 @@ func apply_properties_and_finish() -> void:
 						if property in entity_node:
 							if typeof(entity_node.get(property)) == typeof(properties[property]):
 								entity_node.set(property, properties[property])
-							else:
+							elif not try_handle_implicit_godot_type_conversion(entity_node, property, properties):
 								push_error("Entity %s property \'%s\' type mismatch with matching generated node property." % [entity_node.name, property])
 		
 		if 'func_godot_properties' in entity_node:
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
This PR is no longer compatible with the current version of FuncGodot. Additionally, users should just be making sure their default property types are set appropriately when attempting to use `auto_apply_matching_properties`. FuncGodot will warn them if they're not.

PR #104, opened 2025-05-12
Fix: Trenchbroom RC5 now has the correct orientation for model() classes

RevoluPowered:
I've removed the manual changes to the angles, since the axis conversions are now correct when you swap to RC5 your map will break.

Video showing the issue and the fix https://www.youtube.com/watch?v=t4tWzLAsNVo

You must do the following:
- update func_godot
- update trenchbroom
- go into your map and where you've used model point classes correct the angle

How can I test the assumptions here?
- Import this model into godot and func_godot  (https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/BoomBoxWithAxes) [BoomBoxWithAxes.glb.zip](https://github.com/user-attachments/files/20160204/BoomBoxWithAxes.glb.zip)

- Make a dummy entity with this as the scene file
- Export your FGD, and spawn the entity, it will be oriented 1:1 what it is in godot when you spawn it.
- You can verify this model in blender too and it will do the same.

Why did this work:
- we previously had to compensate for a 90 degree offset on items, there was an additional 180 degrees applied to flip an axis

Map files will need updated due to the manual flip being required once for Model classes.

Upstream trenchbroom issue I created last night:
https://github.com/TrenchBroom/TrenchBroom/issues/4853

This is a breaking change meaning you'll need to inform users of this, it sucks but it should only be needed once. It's why I spent an hour or two just verifying the rotation orders with this model. 

On the left there is a godot instance of the glb file, on the right there is a func godot model entity with the same glb exported and imported from .map
Screenshot 2025-05-12 at 12 24 59

In trenchbroom RC5
Screenshot 2025-05-12 at 12 25 35

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_model_point_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_model_point_class.gd
@@ -97,8 +97,6 @@ func _create_gltf_file(gltf_state: GLTFState, path: String, node: Node3D) -> boo
 	var gltf_document := GLTFDocument.new()
 	gltf_state.create_animations = false
 	
-	node.rotate_y(deg_to_rad(-90))
-	
 	# With TrenchBroom we can specify a scale expression, but for other editors we need to scale our models manually.
 	if target_map_editor != TargetMapEditor.TRENCHBROOM:
 		var scale_factor: Vector3 = Vector3.ONE
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -475,7 +475,6 @@ func build_entity_nodes() -> Array:
 							if not angle is float:
 								angle = float(angle)
 							angles.y += angle
-						angles.y += 180
 						node.rotation_degrees = angles
 					
 					if 'scale' in node and entity_definition.apply_scale_on_map_build:


rtxa:
Does this work for 3D rotations when using mangles/angles?

RevoluPowered:
> Does this work for 3D rotations when using mangles/angles?
angle is tested and works properly for me

can someone please test mangle in RC5?

RhapsodyInGeek:
So this PR won't work for a number of reasons I already shared in the Discord, but I'm going to reiterate them here so there's an easy way to track the conversation.

So this PR doesn't work for 2 major reasons: coordinate system differences and inflexibility. There are also a number of temporary lesser reasons that it can't be merged at this time, even in part.

Tackling inflexibility, this PR assumes the user is using TrenchBroom, but the plugin is intended to be editor agnostic as much as possible. While TrenchBroom recently introduced a fix for GLB coordinate differences in their ASSIMP implementation, we cannot expect other editors that support GLB to follow suit (eg: NetRadiant Custom). The old hardcoded -90 degree turn, while it works generally for Quake coordinate systems, is still inflexible and needs a replacement now for sure. Simply getting rid of it won't do however.

A forward thinking solution for that problem would be to instead add a Vector3 rotation and float scale, to make it flexible enough for any editor's personal implementation of GLB. A Transform3D is also possible, but it should be easy enough to see how excessive that would be. Instead of the hardcoded -90 degree turn, instead we can add the rotation and multiply the scale during GLB generation. This would make it compatible with both TrenchBroom and NetRadiant-Custom, as well as any other untested editors (does Hammer have GLB support?).

As far as removing the 180 degree turn on all Point Entities, that is a non-starter. The whole reason the rotation is necessary is because the forward axis in Quake is positive while forward in Godot is negative along the forward axis. Point Entities != Display Models, and doing this points the actual entities in the wrong direction when -Z is considered forward by Godot.

The topic of "camera forward vs model forward" is an exhaustive one with no true "correct" answer. Blender itself makes both of them opposite: Y is considered camera forward while -Y is considered model forward. It exports GLTF as -Y = +Z. Meanwhile in Godot your camera's forward direction and physics forward direction and transform forward etc etc... has -Z as forward. So your model is imported with it facing the camera, hence why many tutorials mention rotating your model 180 degrees. They did add a "model_forward" implementation, though your mileage may vary with it.

If you're using Model Point Class, then having a rotation property to apply to the exported GLB I think would be sufficient to solve most use cases where you decided to go against this convention.

More reading here: [Godot Docs: Importing 3D Scenes > 3D asset direction conventions](https://docs.godotengine.org/en/4.1/tutorials/assets_pipeline/importing_scenes.html#d-asset-direction-conventions)

RhapsodyInGeek:
Closing this pull request as an alternative fix is desired as well as RC5 being a Pre-Release and therefore not directly supported by FuncGodot (we only support "Latest" releases, not "Pre-Release" versions).

PR #127, opened 2025-07-06
Change map file reading from FileAccess to GDScript.load

suhankins:
My game builds all levels dynamically at runtime, and it ran just fine from the editor, but when exported, apparently FileAccess could not read map files anymore.

This PR removes FileAccess map reading logic and replaces it with GDScript.load instead, making code substantially shorter, easier to understand, and allowing it to work at runtime in my game.

Could be a breaking change, potentially, if anyone is using current FileAccess logic to load maps outside of the project. Should try doing that in my game and see if it works, and if it doesn't, what can be done about it.

### DIFF
--- a/addons/func_godot/func_godot_local_config.tres
+++ b/addons/func_godot/func_godot_local_config.tres
@@ -4,4 +4,3 @@
 
 [resource]
 script = ExtResource("1_g8kqj")
-export_func_godot_settings = false
--- a/addons/func_godot/src/core/func_godot_map_parser.gd
+++ b/addons/func_godot/src/core/func_godot_map_parser.gd
@@ -33,33 +33,16 @@ func load_map(map_file: String, keep_tb_groups: bool) -> bool:
 	component_idx = 0
 	valve_uvs = false
 	_keep_tb_groups = keep_tb_groups
-	
-	var lines: PackedStringArray = []
-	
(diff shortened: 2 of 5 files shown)


RhapsodyInGeek:
> Could be a breaking change, potentially, if anyone is using current FileAccess logic to load maps outside of the project. Should try doing that in my game and see if it works, and if it doesn't, what can be done about it.
 
This PR can't be merged without testing that. We also would need an explanation as to _why_ the fix works, _what_ it actually _does_. Why can't it read map files anymore? Why does load work in this case? If we don't know why we have to fix something, it's likely not a real fix.

My first thought, looking over the old code and looking at your replacement and the issue you're having, is that what you're running into is a recent change (came with 4.4?) they made to resource path properties, where imported resource paths return UIDs now instead of the strings you'd expect. I haven't tested to verify, but I'm almost positive your change breaks external files, which retain their string paths. We appreciate proactiveness, but make sure to test thoroughly before submitting.

That said, in this particular case I wouldn't worry about exploring the issue further for now. We're nearing the end of a complete rewrite of the core scripts, including a parser rewrite that accounts for UIDs which may resolve the issue that you're having. It is something we'll be testing for before releasing the core rewrite update.

Closing PR as a different fix is planned.

PR #129, opened 2025-07-07
prevent unwrapping UVs on children of ModelPoint entite scenes

### LINKED ISSUE #128: unwrap uv2 temporarily breaks Meshinstance3D
In the manual, it says that Unwrap UV2 will cycle through all Solid Entities and unwrap them, but this occures with MeshInstance3D children of ModelPoint Entities as well.

https://github.com/func-godot/func_godot_plugin/blob/ebaace773ffdb7b5fcf331d8b18a17d8b0753979/addons/func_godot/src/map/func_godot_map.gd#L272

unwrap_uv2 will actually unwrap all MeshInstance3Ds that have their GI mode set to static. 

So any ModelPoint Entity who's scene has a MeshInstance3D child with their GI mode set to static will have their UVs messed with, which usually breaks their material until the project is reloaded

I don't think there's a good way to tell if a node was a solid entity, so maybe the docs should just be updated.

benbot:
We shouldn't unwrap the UVs of MeshInstances of ModelPoint Entities. It ends up breaking lightmapping unless you do a really weird specific unwrap -> reload project -> bake sequence.

Should resolve #128

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -269,7 +269,7 @@ func unwrap_uv2(node: Node = null) -> void:
 		print("Unwrapping mesh UV2s")
 	
 	if target_node is MeshInstance3D:
-		if target_node.gi_mode == GeometryInstance3D.GI_MODE_STATIC:
+		if target_node.gi_mode == GeometryInstance3D.GI_MODE_STATIC and target_node.owner == owner:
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
We'll actually be using a completely different way of unwrapping UV2s on meshes in the upcoming Core Rewrite, so this fix is a moot point. Feel free to keep it in your personal project as a bandaid for now, but the rewrite is getting closer to release.

PR #133, opened 2025-08-07
Build_visuals solid class definition fix

RisingThumb:
FuncGodotFGDSolidClass has a build_visuals property which is not currently being taken into account in the rewrite. This is a small fix that checks this property before building a MeshInstance3D


sinewavey:
Good catch! A fix should be out shortly

PR #136, opened 2025-08-15
Bug fixes for mesh metadata generation

LogicAndTrick:
- Move texture metadata to the mesh building loop to keep all metadata logic in the same place
- Change normals metadata to face normals, rather than vertex normals from the mesh builder
- Use the existing get_centroid function to calculate face positions
- Add vertices using the triangle indices rather than the vertex array from the mesh builder
- Also a tiny whitespace fix on the comments at the end - they were indented too much

To test this I made a simple map with 5 visible quads using 4 different textures. Expectation is  that texture_names has 4 entries, textures/normals/positions all have 5, and vertices has 30 (5 quads -> 10 tris -> 30 verts).

Before:
image

After:
image

### DIFF
--- a/addons/func_godot/src/core/geometry_generator.gd
+++ b/addons/func_godot/src/core/geometry_generator.gd
@@ -321,13 +321,6 @@ func generate_entity_surfaces(entity_index: int) -> void:
 			if not surfaces.has(face.texture):
 				surfaces[face.texture] = []
 			surfaces[face.texture].append(face)
-			
-			if def.add_textures_metadata:
-				var tex_index: int = texture_names_metadata.find(face.texture)
-				if tex_index < 0:
-					tex_index = texture_names_metadata.size()
-					texture_names_metadata.append(face.texture)
-				textures_metadata.append(tex_index)
 	
 	# Cache order for consistency when rebuilding 
 	var textures: Array[String] = surfaces.keys();
@@ -346,6 +339,11 @@ func generate_entity_surfaces(entity_index: int) -> void:
 	for texture_name in textures:
 		# SURFACE SCOPE BEGIN
 		faces = surfaces[texture_name]
+		
+		# Get texture index for metadata
+		var tex_index: int = texture_names_metadata.size()
+		if def.add_textures_metadata:
+			texture_names_metadata.append(texture_name)
 
 		# Prepare new array
 		arrays = Array()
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
So two issues I can find. First issue is that your centroids are pre-scaled, so you get the actual Quake units rather than Godot measurements. Additionally they're pre-converted, so you're working in Quake coordinates rather than Godot coordinates. There is an `id_to_gl()` util function that can be used to convert the positions and normals to the Godot coordinate system. The FuncGodotMap's Map Settings also contains a scale factor property that can be multiplied against the centroids and vertices to get the proper Godot measurements.

How I tested it in test_project:
```gdscript
@tool
extends Node

@export_tool_button("Show Face Data") var show_face_data : Callable = _show_face_data

func _show_face_data() -> void:
	var fgm := $FuncGodotMap
	var scene_root = fgm.owner
	
	for c in fgm.get_children():
		if c.has_meta("func_godot_mesh_data"):
			var mdata: Dictionary = c.get_meta("func_godot_mesh_data", Dictionary())
			var t: PackedInt32Array = mdata["textures"]
			var tn: Array = mdata["texture_names"]
			var v: PackedVector3Array = mdata["vertices"]
			var n: PackedVector3Array = mdata["normals"]
			var p: PackedVector3Array = mdata["positions"]
			
			for i in t.size():
				var label := Label3D.new()
				label.text = tn[t[i]]
				label.font_size = 16
				c.add_child(label)
				label.owner = scene_root
				label.position = p[i] * (fgm as FuncGodotMap).map_settings.scale_factor + n[i] * -0.1
```

Result:
image

I think there are other things going on too, but that's what I've been able to figure out so far.

Here's the project I was testing with:
[mesh_data_test_project.zip](https://github.com/user-attachments/files/21800279/mesh_data_test_project.zip)

LogicAndTrick:
Thanks, that makes sense. I've updated the implementation to use triangle indices instead of face indices as well as properly converting the positions to Godot coordinates.

Using your test project and changing the last line in the `_show_face_data()` function to `label.position = p[i] + n[i] * 0.1`, it looks like this now:
image

sinewavey:
Yeah, this is likely related to the distance_squared based check in build_entity_vertices for merging. I was trying (albeit, unsuccessfully so far) to find a better approach with spatial hashing to reuse and merge close-enough vertices. It's also likely what led to some issues with the new implementation for smooth shading in geo generator itself, too.

LogicAndTrick:
I recommend using the winding + clipping approach to generating the faces instead of using the 3-plane intersection method - I've found it to be much better in regards to rounding errors, and is an easier to understand algorithm on top of that.

Basically the approach is what quake compilers use - make a huge face from the plane, and then clip it against every other plane in the solid, discarding the front side. The final result is the face polygon. Doing those calculations with doubles and only switching to floats at the end helps a lot too.

sinewavey:
fortunately with how the refactor was done this should be something that is able to be swapped in and out relatively pain free -- definitely on the list of things to investigate this upcoming week!

sinewavey:
Hi, just a quick update: implementing the hypersquare approach was a lot easier (and even cleaner and shorter than the existing method). So far, it seems to be generating fewer vertices overall, as expected.

I'll be doing some more testing later today on more complex maps and brushes, but hopefully should have a release featuring this shortly.

The approach also winds faces as part of the clipping faces (assuming Geometry3D::clip_polygon() preserves winding order), which is another huge bonus.

Thank you for your suggestion. I'm initially working with a base face polygon size of 65,535.0, but this seems to be an arbitrary size that might need to be exposed ultimately (though I doubt)

RhapsodyInGeek:
@LogicAndTrick can you merge the latest codebase into your branch and make sure that your fixes are still working? We recently merged the hypersquare and tangent fixes, which should have some sort of impact on the mesh data collection.

RhapsodyInGeek:
I'll give it a test run myself and probably merge it today then.

PR #137, opened 2025-08-17
Fix for string default values in 'choices' properties

LogicAndTrick:
The documentation and code allows both String and int types as values for choices properties, however when exporting the FGD the value is not quoted, causing parsing errors when the value is of string type.

The fix just quotes the value, which works for both string and int types.


LogicAndTrick:
Fair enough, updated!

RhapsodyInGeek:
Looks good!

PR #138, opened 2025-08-17
Export bool property defaults to FGD

LogicAndTrick:
Default values for boolean properties are fully supported, but their default value is not added to the FGD so it is not visible in the editor.

This change would update the FGD from:

```
name(choices) : "description" = 
[
	0 : "No"
	1 : "Yes"
]
```

to:

```
name(choices) : "description" : 1 = 
[
	0 : "No"
	1 : "Yes"
]
```


RhapsodyInGeek:
Looks good!

PR #140, opened 2025-08-18
Add check for clip brushes to generate_entity_surfaces

suhankins:
Right now, any map that has `special/clip` will crash with error `res://addons/func_godot/src/core/geometry_generator.gd:478 - Invalid access to property or key 'special/clip' on a base object of type 'Dictionary[String, Material]'.`

With this change, `special/clip` should no longer get added to `textures_metadata`.


RhapsodyInGeek:
Closing this PR in favor of a more suitable solution.

PR #148, opened 2025-08-26
Resolve UID path in verify()

suhankins:
Right now, if your game passes map file to FuncGodotMap by UID, during building it will crash with errors

```
ERROR: Invalid UID.
   at: (core/io/resource_uid.cpp:154)
ERROR: [MAP] Map file uid://SOMEUID does not exist.
   at: push_error (core/variant/variant_utility.cpp:1098)
ERROR: [MAP] Verification failed: Does not exist. Aborting map build
   at: push_error (core/variant/variant_utility.cpp:1098)
```

Resolving UID to path in `verify()` resolves this issue.

Personally I'm not even sure why is `verify()` needed, considering `parse_map_data()` in `parser.gd` already verifies itself that everything is correct and map file exists.

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -65,6 +65,14 @@ func verify() -> Error:
 		fail_build("Cannot build empty map file.")
 		return ERR_INVALID_PARAMETER
 	
+	# Retrieve real path if needed
+	if _map_file_internal.begins_with("uid://"):
+		var uid := ResourceUID.text_to_id(_map_file_internal)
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
Looks good to me!

> Personally I'm not even sure why is verify() needed, considering parse_map_data() in parser.gd already verifies itself that everything is correct and map file exists.

There are probably some redundancies here and there in the code that could be gotten rid of. The rewrite was developed mainly by 2 people and there were *a lot* of components to making it work, and clearly more than a few missed issues. We'll be gradually going over the code to clean it up more as time goes on.

PR #149, opened 2025-08-27
Implement the collision shape to face indices map metadata

LogicAndTrick:
This re-implements the functionality of the `collision_shape_to_face_range_map` metadata. Since the faces for each brush are no longer contiguous within a mesh, I've changed the metadata type from a range to a list of indices, and also changed the name of the metadata key to match. The configuration variable name should probably be updated to suit the change in data type.

## Geometry generator
For each face, I store a dictionary to map faces to the list of metadata indices for each triangle of the face. Then, when the collision shapes are generated:

- Convex shapes: when creating the collision shape for a brush, we loop through each face of and collect the metadata indices associated with that brush
- Concave shapes: only one shape is generated per entity, so we can simply collect the whole list of metadata indices associated with the entity (This means the resulting metadata will simply be an array from 0 to num_tris in the mesh, making this metadata type of little use when using concave collision. However, it is implemented here for consistency across both collision types.)

Either way, the data is stored in the `mesh_metadata` to be transformed in the next step.

## Entity assembler
Here I merged the shape creation into a single loop instead of the similar branches based on the concave/convex setting - now it just changes the node name based on the setting but otherwise uses the same logic. An immediate benefit is consistency - for the concave case, it was setting the `owner`, but it wasn't for the convex case. (I assume owner should be set in both scenarios, but let me know if that's wrong)

The other additions simply collects the metadata into a single object, and then attaches the metadata to the entity node.

## Questions
- I've changed the metadata name from `collision_shape_to_face_range_map` to `collision_shape_to_face_indices_map`, is this name fine or should I change it to something else?
- I haven't changed the user-facing `FuncGodotFGDSolidClass` property `add_collision_shape_face_range_metadata` in this PR. Should it get changed at the same time to something more accurate now that the data has changed from a range of indices to a list of indices?

### DIFF
--- a/addons/func_godot/fgd/func_detail.tres
+++ b/addons/func_godot/fgd/func_detail.tres
@@ -22,6 +22,7 @@ add_textures_metadata = false
 add_vertex_metadata = false
 add_face_position_metadata = false
 add_face_normal_metadata = false
+add_collision_shape_to_face_indices_metadata = false
 add_collision_shape_face_range_metadata = false
 classname = "func_detail"
 description = "Static collidable geometry. Builds a StaticBody3D with a MeshInstance3D and a single concave CollisionShape3D. Does not occlude other VisualInstance3D nodes."
--- a/addons/func_godot/fgd/func_detail_illusionary.tres
+++ b/addons/func_godot/fgd/func_detail_illusionary.tres
@@ -22,6 +22,7 @@ add_textures_metadata = false
 add_vertex_metadata = false
 add_face_position_metadata = false
 add_face_normal_metadata = false
+add_collision_shape_to_face_indices_metadata = false
 add_collision_shape_face_range_metadata = false
 classname = "func_detail_illusionary"
 description = "Static geometry with no collision. Builds a Node3D with a MeshInstance3D. Does not occlude other VisualInstance3D nodes."
--- a/addons/func_godot/fgd/func_geo.tres
+++ b/addons/func_godot/fgd/func_geo.tres
@@ -22,6 +22,7 @@ add_textures_metadata = false
 add_vertex_metadata = false
 add_face_position_metadata = false
 add_face_normal_metadata = false
+add_collision_shape_to_face_indices_metadata = false
 add_collision_shape_face_range_metadata = false
 classname = "func_geo"
 description = "Static collidable geometry. Builds a StaticBody3D with a MeshInstance3D, a single concave CollisionShape3D, and an OccluderInstance3D."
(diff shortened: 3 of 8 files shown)


LogicAndTrick:
Sure, here's my code that makes use of this. My player controller has a `ShapeCast3D` at its feet that checks for (floor) surfaces it collides with. Since I can get the collider and the collider shape out, I can use those plus the metadata to find which visual texture the player is walking on (when using convex collision):

```gdscript
# in a CharacterBody3D with a ShapeCast3D at its feet

func determine_surface_texture() -> String:
	var cast := %SurfaceShapecast as ShapeCast3D
	if not is_on_floor() or not cast.is_colliding():
		return ''
	for i in range(cast.get_collision_count()):
		# get collision data
		var normal := cast.get_collision_normal(i)
		var target := cast.get_collider(i) as CollisionObject3D
		var shape_id := cast.get_collider_shape(i)
		var owner_id := target.shape_find_owner(shape_id)
		var shape := target.shape_owner_get_owner(owner_id) as CollisionShape3D
		# make sure the metadata exists on the entity and grab the metadata
		if not target.has_meta('func_godot_mesh_data'):
			continue
		var mesh_meta : Dictionary = target.get_meta('func_godot_mesh_data')
		if not mesh_meta.has('normals') \
			or not mesh_meta.has('textures') \
			or not mesh_meta.has('texture_names') \
			or not mesh_meta.has('collision_shape_to_face_indices_map'):
			continue
		var collision_shape_to_face_indices_map : Dictionary[String, PackedInt32Array] = mesh_meta.get('collision_shape_to_face_indices_map')
		if not collision_shape_to_face_indices_map.has(shape.name):
			continue
		var normals : PackedVector3Array = mesh_meta.get('normals')
		var textures : PackedInt32Array = mesh_meta.get('textures')
		var texture_names : Array[StringName] = mesh_meta.get('texture_names')
		var face_indexes : PackedInt32Array = collision_shape_to_face_indices_map.get(shape.name)
		# face_indexes contains which visual surfaces correspond to this collision shape
		# find the one with approximately the same normal
		for idx in face_indexes:
			var surf_normal : Vector3 = normals[idx];
			if Math.approx_vector3(surf_normal, normal, 0.01):
				return texture_names[textures[idx]]
	return ''
```

Then I can use the texture name to choose and play different footstep sounds:

https://github.com/user-attachments/assets/bc75cc4e-5de4-4b76-9f10-fb101a76a773

LogicAndTrick:
Done! It looks a little noisy in the editor since both names are fairly long, but it works.
image

RhapsodyInGeek:
Looks good!

PR #163, opened 2025-09-06
Unwrap UV2 on smooth shaded meshes after smoothing process

### LINKED ISSUE #162: Entities with phong enabled do not build UV2
Create two func_details, enable phong on one but not the other, and build the map. The entity with phong enabled will not unwrap the UV2 layer, the one without should build as normal.

Hate to be the one bringing up big issues (sorry!), but a little important for lightmaps as you might imagine!

sinewavey:
This PR adds a few small changes to the UV2 unwrap process.

Currently, mesh unwrap occurs in geometry_generator. However, if a mesh is smooth shaded, this is seemingly discarded by the new ArrayMesh returned by `smooth_mesh_by_angle`. So, when a user desired UV2 unwrapping, only non-smoothed meshes would be unwrapped.

This fix should be reverted when smooth shading is moved back into Geometry Generator, as the unwrap would occur after that process already. See the notes in code for more explanation.

As such, if a mesh is detected as using non-disabled global illumination and UV2 unwrap is enabled, it is skipped in GeometryGenerator and instead unwrapped after smoothing in EntityAssembler.

To address typing, `smooth_mesh_by_angle` now returns an ArrayMesh (no code changes, as this is already what was returned)

Should resolve #162 

cc: @RhapsodyInGeek regarding lightmap_unwrap's transform argument - [according to the Godot source code](https://github.com/godotengine/godot/blob/3c7f9b937214068bc892be0d2bd9a8a7026edae1/scene/resources/mesh.cpp#L2083) the lightmap unwrap transform input only reads the scale so I believe this is still in line with what's expected rather than inputting a 3D transform representing the object's position but wanted your thoughts

### DIFF
--- a/addons/func_godot/src/core/data.gd
+++ b/addons/func_godot/src/core/data.gd
@@ -154,6 +154,12 @@ class EntityData extends RefCounted:
 		return (definition
 				and definition is FuncGodotFGDSolidClass
 				and definition.build_visuals)
+
+	func is_gi_enabled() -> bool:
+		return (definition
+				and definition is FuncGodotFGDSolidClass
+				and definition.global_illumination_mode
+		)
 	
 	## Checks the entity's FGD resource definition, returning whether the Solid Class CollisionShapeType is set to Convex.
 	func is_collision_convex() -> bool:
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -95,8 +95,16 @@ func generate_solid_entity_node(node: Node, node_name: String, data: _EntityData
 			node.add_child(occluder_instance)
 			data.occluder_instance = occluder_instance
 		
+		# NOTE: Currently occuring in EntityAssembler until the appropriate method in GeometryGenerator is resolved
+		# For now, smooth entire mesh, then unwrap for lightmap if needed
 		if not (build_flags & FuncGodotMap.BuildFlags.DISABLE_SMOOTHING) and data.is_smooth_shaded(map_settings.entity_smoothing_property):
 			mesh_instance.mesh = FuncGodotUtil.smooth_mesh_by_angle(data.mesh, data.get_smoothing_angle(map_settings.entity_smoothing_angle_property))
+
+			if data.is_gi_enabled() and (build_flags & FuncGodotMap.BuildFlags.UNWRAP_UV2):
+				mesh_instance.mesh.lightmap_unwrap(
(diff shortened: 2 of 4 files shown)


RhapsodyInGeek:
> regarding lightmap_unwrap's transform argument - [according to the Godot source code](https://github.com/godotengine/godot/blob/3c7f9b937214068bc892be0d2bd9a8a7026edae1/scene/resources/mesh.cpp#L2083) the lightmap unwrap transform input only reads the scale so I believe this is still in line with what's expected rather than inputting a 3D transform representing the object's position but wanted your thoughts

Yeah, I remember seeing that before too, though I forget why I was looking it up. I thought it was kind of funny, because it goes out of its way to discard everything except scale. I wonder if it's done that way for potential future proofing?

> This fix should be reverted when smooth shading is moved back into Geometry Generator, as the unwrap would occur after that process already. See the notes in code for more explanation.

Does this mean we're not planning on keeping this fix? Is there any way we can keep it in Geometry Generator if that's where we ultimately want it to end up being? Is the mesh resource being altered in Entity Assembler for some reason? I'm pretty positive I put it there for some reason, which I think may have been related to Map Settings not being passed to Geometry Generator originally.

Looking over the code, I don't see any reason we can't put the mesh smoothing from Entity Assembler into Geometry Generator now, just before the UV unwrap occurs, since it should have the same effect on performance.

sinewavey:
> Does this mean we're not planning on keeping this fix? Is there any way we can keep it in Geometry Generator if that's where we ultimately want it to end up being? 

Ultimately, the fix is a problem of order of operations. Ideally, we wouldn't do smoothing on the mesh after geometry generation; it could be done during. Though the skeleton code process in place now wasn't working correctly - perhaps after revisiting with vertex merging it may function a bit better.

> Is the mesh resource being altered in Entity Assembler for some reason? I'm pretty positive I put it there for some reason, which I think may have been related to Map Settings not being passed to Geometry Generator originally.

The mesh resource gets replaced entirely in Entity Assembler, as that's where smoothing is done currently, because of the following point:
 
> Looking over the code, I don't see any reason we can't put the mesh smoothing from Entity Assembler into Geometry Generator now, just before the UV unwrap occurs, since it should have the same effect on performance.

When I tried this with the current implementation of smooth_mesh_by_angle, it would cause engine crashes in Geo Gen. I tried a number of solutions, including forcing it to be sequential. The mesh would either be null going in (somehow) or a total lockup would happen.

PR #167, opened 2025-10-04
Remove Argument from FileAccess.get_text() call in Quake Map File Importer

plaught-armor:
4.6 doesn't require parameter to be passed to `get_as_text()`


plaught-armor:
> We don't officially support 4.6 yet. We only support the latest stable release.
> 
> We also need more in your description than just what the fix does. Why is this needed? What does it fix? What was the problem the plugin was having? And where did you hear 4.6 doesn't require the parameter to be passed? According to the docs, if I want it set to `true` I have to pass a parameter as it defaults to `false`. https://docs.godotengine.org/en/latest/classes/class_fileaccess.html#class-fileaccess-method-get-as-text
> 
> I'm trying to remember why I had set it to ignore carriage returns, as I do recall intentionally setting it to true for a reason. I'd have to look it up again though, as the reason escapes me at the moment.


image

I guess if setting it to true is required then something may need to be worked around. I guess this is just something to look out for.

RhapsodyInGeek:
I'm going to leave this PR open for now and link to this discussion they were having on it. Seems that they decided to remove it due to considering the method overscoped? In any case, it'll be helpful to have this remain open for reference when 4.6 goes stable.

https://github.com/godotengine/godot/pull/110867

PR #170, opened 2025-10-12
Made generated model size respect default inverse scale factor

benbot:
The docstring for the model point's `scale_factor` field says

> Scale expression applied to model. Only used by TrenchBroom. If left empty, uses [member ProjectSettings.func_godot/default_inverse_scale_factor]. Read the TrenchBroom Manual for more information on the "scale expression" feature.

But when generating the glbs, if no scale expression is set, then it does not use the default inverse scale factor on the generated model.

This change fixes that. So now no scale expression will be the same as the project wide inverse scale factor.


RhapsodyInGeek:
Closing as this was already fixed:

https://github.com/func-godot/func_godot_plugin/blob/main/addons/func_godot/src/fgd/func_godot_fgd_model_point_class.gd#L174

RhapsodyInGeek:
Ah, my mistake, it was fixed for non TB map editors. Looks good to me!

Issue #178, opened 2025-10-23
Improve use Generated Materials for Shader Materials


RisingThumb:
Whenever you use a generated material and your default material is a shader material, the current behaviour whenever a material is generated, is to make a deep copy of that material. This works well for the shader parameters, however for the shader itself, it means if you are using a global shader, you have to go through every material to change it to a reference to that shader rather than a unique copy of it, otherwise edits made to the global shader will not be reflected in your maps until you change the material. If you have a lot of textures this becomes a source of a lot of frustration

I personally think default behaviour should be that the shader is not a unique copy(I have not encountered a situation where this behaviour has been beneficial to me, I am open to hearing if it's beneficial to others?). This could be implemented as a toggle in Map Settings for making shaders unique.

Additionally, as we have the Albedo Map Pattern, Normal Map Pattern etc provided for automatically generated PBR materials, perhaps it would also be nice to also allow shader uniforms to go alongside these too for setting them on ShaderMaterials? The current approach only works for anything inheriting [BaseMaterial3D with set_texture](https://docs.godotengine.org/en/stable/classes/class_basematerial3d.html#class-basematerial3d-method-set-texture). I know for my usecase, having a uniform for normal maps would be very useful

PR #180, opened 2025-10-23
Add option in Map Settings for using a unique shader for generated materials

RisingThumb:
Addresses first half of #178 . Makes things a lot less tedious if you have a lot of textures all sharing a common shader.

For now I've defaulted it to false(meaning it will not generate a unique shader). I think this is a sensible default because this should prevent mistakes writing shaders(writing the shader to a generated material and not the global and saved shader file), and maybe avoid some shader compilation costs as it's all a common shader(though I'm not as sure on this 🤔)

### DIFF
--- a/addons/func_godot/src/map/func_godot_map_settings.gd
+++ b/addons/func_godot/src/map/func_godot_map_settings.gd
@@ -114,10 +114,6 @@ var scale_factor: float = 0.03125
 ## [i]NOTE: Materials do not use the [member default_material] settings after saving.[/i]
 @export var save_generated_materials: bool = true
 
-## When saving a generated material, whether to make a unique shader or use a reference to the shader.
-## [i]NOTE: Only relevant if [member save_generated_materials] is true and [member default_material] is a Shader Material.[/i]
-@export var unique_shader: bool = false
-
 #endregion
 
 @export_category("UV Unwrap")
--- a/addons/func_godot/src/util/func_godot_util.gd
+++ b/addons/func_godot/src/util/func_godot_util.gd
@@ -202,9 +202,7 @@ static func build_texture_map(entity_data: Array[FuncGodotData.EntityData], map_
 				
 				# Material generation
 				elif map_settings.default_material:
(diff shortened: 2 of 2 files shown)


RhapsodyInGeek:
From #178 
> Whenever you use a generated material and your default material is a shader material, the current behaviour whenever a material is generated, is to make a deep copy of that material. This works well for the shader parameters, however for the shader itself, it means if you are using a global shader, you have to go through every material to change it to a reference to that shader rather than a unique copy of it, otherwise edits made to the global shader will not be reflected in your maps until you change the material. If you have a lot of textures this becomes a source of a lot of frustration

Looking at the line
```gdscript
var material = map_settings.default_material.duplicate(true)
```
and realizing that's honestly a mistake on my part. It isn't intended that the material duplicate the shader, just the default material itself. So a proper fix would be to turn off deep copy, right?

I can't think of any reasons why we would want to make shaders unique off the top of my head?

RisingThumb:
I don't really see much point in making shaders unique, I agree

> and realizing that's honestly a mistake on my part. It isn't intended that the material duplicate the shader, just the default material itself. So a proper fix would be to turn off deep copy, right?

It would fix this issue. I think it still makes sense to have a deep copy of a ShaderMaterial still(aside from the Shader itself). It's useful for NoiseTextures that should be deep copies(and some other textures that map to a Sampler2D probably make sense too), and you can easily do common uniforms across shaders already with [global shader uniforms](https://docs.godotengine.org/en/stable/tutorials/shaders/shader_reference/shading_language.html#doc-shading-language-global-uniforms)

RhapsodyInGeek:
> I think it still makes sense to have a deep copy of a ShaderMaterial still

Map Settings wouldn't be the place to do it though, as that'd be an "all or nothing" type deal which would be inappropriate for something like this. I don't know how we'd do it on a per material basis though, so I think I'd just leave details like that up to the developer to solve (something like a KVP on an entity using that material that they could apply when needed, which would be outside func_godot's scope).

PR #185, opened 2025-10-30
Add optional partitioning of solid class meshes

zyebytevt:
This might be a bit of a naive implementation so feedback is welcome, but what this change does is it adds a `partition_grid_size` property to `FuncGodotFGDSolidClass`, and if such is greater than `Vector3.ZERO`, brushes will be split into multiple meshes and occluders, depending on the center position of it's faces.

This is mainly for use with worldspawn, so that one can map as usual without having to separate everything into func_geo's without the associated performance impact on huge scenes.

### DIFF
--- a/addons/func_godot/src/core/data.gd
+++ b/addons/func_godot/src/core/data.gd
@@ -135,9 +135,9 @@ class EntityData extends RefCounted:
 	## This can only be a [FuncGodotFGDSolidClass], [FuncGodotFGDPointClass], or [FuncGodotFGDModelPointClass].
 	var definition: FuncGodotFGDEntityClass = null
 	## Mesh resource generated during the geometry generation stage and applied during the entity assembly stage.
-	var mesh: ArrayMesh = null
+	var meshes: Array[ArrayMesh] = []
 	## MeshInstance3D node generated during the entity assembly stage.
-	var mesh_instance: MeshInstance3D = null
+	var mesh_instances: Array[MeshInstance3D] = []
 	## Optional mesh metadata compiled during the geometry generation stage, used to determine face information from collision.
 	var mesh_metadata: Dictionary = {}
 	## A collection of collision shape resources generated during the geometry generation stage and applied during the entity assembly stage.
@@ -145,7 +145,7 @@ class EntityData extends RefCounted:
 	## A collection of [CollisionShape3D] nodes generated during the entity assembly stage. Each node corresponds to a shape in the [member shapes] array.
 	var collision_shapes: Array[CollisionShape3D] = []
 	## [OccluderInstance3D] node generated during the entity assembly stage using the [member mesh] resource.
-	var occluder_instance: OccluderInstance3D = null
+	var occluder_instances: Array[OccluderInstance3D] = []
 	## True global position of the entity's generated node that the mesh's vertices are offset by during the geometry generation stage.
 	var origin: Vector3 = Vector3.ZERO
 
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -61,26 +61,27 @@ func generate_solid_entity_node(node: Node, node_name: String, data: _EntityData
 	var properties: Dictionary = data.properties
 	
 	# Mesh Instance generation
-	if data.mesh:
+	var mesh_instance_idx := 0
+	for mesh: Mesh in data.meshes:
 		var mesh_instance := MeshInstance3D.new()
-		mesh_instance.name = node_name + "_mesh_instance"
-		mesh_instance.mesh = data.mesh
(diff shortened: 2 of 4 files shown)


RhapsodyInGeek:
While this implementation is neat, it won't become part of the official plugin. It doesn't really help that much with optimization, as you want your occluders to be as simple as possible and if you try to build all of your level geometry as Worldspawn with occluders enabled on it, you're probably going to run into issues with over-complicated occluders. 

Echoing RisingThumb, what happens with spheres and arches and unclean geometry? When you chop them up indiscriminately, you can create more optimization issues than you might've had if you just left them alone. You can easily run into unnecessarily excessive draw calls this way, too.

Another thing to consider is how Godot's Physics systems attempt to optimize for collisions. It's not an area I have complete knowledge in, so someone correct me if I'm wrong, but as far as I understand you can optimize collision polling more by having multiple StaticBody3D nodes, as there is a broad phase pass that can exclude them from later checks.

Another collision issue you can run into is the seams between different objects. There can be a noticeable "hiccup" when trying to walk across collision seams. If you just have it chopping up the collision indiscriminately based on a grid, then you run into regular hiccups when walking across even flat expanses.

It also takes func_godot away from being a "what you see/make is what you get". Sure, you *could* try to predict where certain grid cuts would happen to avoid the aforementioned issues... but now you're *really* building on a grid. That seems like just as much work to make something good as it would've been to just manually split your entities up as you go along.

I know it can be frustrating that you can't map like you would with Quake using func_godot, and a lot of folks want shortcuts that can allow them to do so, but sometimes you really do need to just "do the work".

That's not to say all "shortcuts" are bad. A good example is #182 from RisingThumb, where it func_godot can attempt to determine what faces will never be seen. Being able to do that on an entity by entity basis can be a really helpful QOL feature.

That all said, if this change helps you in your personal projects, by all means keep exploring it. A lot of people have their own personal forks of func_godot with features that their developers find useful, even if they don't match the vision or intent of the plugin.

I'll leave this PR open for discussion for now, as it's still an interesting idea to explore.

RhapsodyInGeek:
> Just to note this also happens when you split a brush up using func_geo as the docs suggest. To get around it you have to basically have 2 brushes, one to be the mesh you split up as func_detail_illusionary and one to be the collision as a clip brush. Doing this starting with a huge amount of worldspawn is a huge time sink, and liable to run into human error.

My point is that when manually create your rooms + areas, you have much more control over these hiccups by manually determining how rooms connect with one another so that you can minimize the effect of connected colliders much more naturally.

> which still applies when you break things up with func_geo... probably with more MeshInstance3Ds created doing it manually due to human error

> but if this is seriously robust and in the hands of a competent developer

It's just as easy to say that you would have fewer MeshInstance3Ds created doing it manually due to developer experience and that this feature would be seriously detrimental in the hands of an incompetent developer. Both perspectives would be valid assessments.

> it will seriously cut down the time spent on the optimisation step by splitting things up with func_geo

It really doesn't take all that much time to turn a group of brushes into a func_geo as you go, and you should probably be using func_detail more than func_geo anyway so that you create fewer + simpler occluders while creating flexible cullable instances. I have never been sold on the idea that creating brush entities is time consuming. It really is just a matter of "update your work flow to match with what Godot wants, not Quake".

What can actually be time consuming is "Skipping" interior faces. Sometimes even having a good work flow for it won't save you from the potential tedium. Incidentally, having your level split up entirely into the various func entities actually makes this process easier.

RisingThumb:
> To be clear, I was specifically talking about #182 when I said "Being able to do that on an entity by entity basis can be a really helpful QOL feature."

I know but imo it's the right approach here as you get control over which brush entities get partitioned, meaning if weird geo behaves poorly with it, you can make it not be partitioned. Anyways goes back to wanting to see how robust it is

> It's just as easy to say that you would have fewer MeshInstance3Ds created doing it manually due to developer experience and that this feature would be seriously detrimental in the hands of an incompetent developer. Both perspectives would be valid assessments

This is also not something I'm 100% on though. I know Godot batches MeshInstance3Ds for draw calls, so an extra MeshInstance3D doesn't immediately equal another draw call. If you know more about this, can you point me in the right direction to understand it better?

I kinda agree, but I'm wary about safeguarding the incompetent developer at the cost of what a competent developer is allowed to use

>  I have never been sold on the idea that creating brush entities is time consuming. 

Do you mean just clicking brushes to make brush entities, or do you mean the steps involved that this PR attempts to solve with chopping brushes up and making new brush entities and skipping the invisible faces, and making a clip brush if it's somewhere that causes collision hiccups, and then also setting it up for occluders if necessary? All the steps in the latter add up and become very time consuming. You mentioned skipping interior faces being time consuming, but if this PR just chops meshes up and doesn't create any new faces, you have no new interior faces to deal with. I reckon the time saved is at least in the tens of minutes, but that's just a guess, I'd have to use it to find out

Issue #186, opened 2025-11-03
Using `Image` import options causes `load_texture` to crash


ambiguousname:
`load_texture` does not validate that a given texture is of type `Texture2D`, leading to a crash when `load_texture` returns a `Texture2D` type that is actually an `Image`. Then when `texture.get_size()` is called, this leads to a memory leak and a subsequent crash.

For example, let's say I have a file `textures/white.webp` that I have imported as an `Image`, that is used as a texture in my `.map` file. Then in `build_texture_map`:

```gdscript
if not texture_sizes.has(texture_name):
	var texture: Texture2D = load_texture(texture_name, wad_resources, map_settings)
	if texture:
		texture_sizes[texture_name] = texture.get_size()
```

In `load_texture`, `textures/white.webp` is recognized as a valid texture and is then `load`ed. Godot in this case doesn't seem to realize that the resource it has loaded is not a valid `Texture2D` on a return and so this leads to the memory issues. I'm working on a PR to just return `null` if `load_texture` cannot validate correctly.

I understand that this is kind of a weird edge case; I'm using a custom shading solution and am only interested in using textures to indicate how something should be shaded, not to add actual texture information.

PR #187, opened 2025-11-03
Suggested fix for #186

ambiguousname:
Quick fix for #186 that fixes my crashing issues.


RhapsodyInGeek:
This needs a lot more information. You need to describe the issue and the solution and why it's needed. Just linking to the issue is not enough, even if you end up repeating much of what you described in the issue.

PR #194, opened 2025-11-07
Fix node_class check in entity assembler to support all script based classes

avetharun:
This allows FuncGodot's map builder to use CSharpScripts with the node_class property on the base FuncGodotFGDEntityClass so it keeps the same typing and usage as with script_class when using Point/Solid classes.

This fix allows people to use CSharpScripts where GDScript was required in node_class

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -45,7 +45,7 @@ var prefix: String = ""
 
 @export_group("Node Generation")
 
-## Node to generate on map build. This can be a built-in Godot class, a GDScript/CSharpScript class, or a GDExtension class. 
+## Node to generate on map build. This can be a built-in Godot class, a Script class, or a GDExtension class. 
 ## For Point Class entities that use Scene File instantiation leave this blank.
 @export var node_class := ""
 


avetharun:
This patch allows FuncGodot's map builder to use CSharpScripts with the `node_class` property on the base FuncGodotFGDEntityClass so it keeps the same typing and usage as with `script_class` when using Point/Solid classes.
I much prefer to write my games in C# rather than GDScript and find it easier to use node_class when I have a project with >100 CSharpScript files for the `script_class` property

Basically, this fix allows people to use CSharpScripts where GDScript was required in `node_class`

I also apologize for being awful at writing PRs, its a bad habit.

addons/func_godot/src/core/entity_assembler.gd
```diff
@@ -51,8 +51,8 @@ func generate_solid_entity_node(node: Node, node_name: String, data: _EntityData
 				node = ClassDB.instantiate(definition.node_class)
 		else:
 			var script: Script = get_script_by_class_name(definition.node_class)
-			if script is GDScript:
-				node = (script as GDScript).new()
+			if script != null and script is Script:
```
RhapsodyInGeek: Do we need to check for `null` here? As far as I understand, this would be redundant as `if Script` returns false if `script == null`.

avetharun: True, initially just used != false to check, but decided on Script

RhapsodyInGeek:
Looking into it more, I believe we specified GDScript because we wanted to mitigate malicious behavior with the script system. I would say change the condition to
```gdscript
if script is GDScript or script is CSharpScript:
```
but CSharpScript is not available in the non-C# versions of Godot, and so we wouldn't be able to support that.

Maybe @sinewavey can remember our discussions on the safety of allowing any script to act as a class, or if it really was just an oversight? Maybe it's much ado about nothing.

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
```diff
@@ -45,7 +45,7 @@ var prefix: String = ""
 
 @export_group("Node Generation")
 
-## Node to generate on map build. This can be a built-in Godot class, a GDScript class, or a GDExtension class. 
+## Node to generate on map build. This can be a built-in Godot class, a GDScript/CSharpScript class, or a GDExtension class.
```
RhapsodyInGeek: Talking it over with the team, this would more accurately be `Node to generate on map build. This can be a built-in Godot class, a Script class, or a GDExtension class.` since we would be allowing *any* kinds of scripts inheriting from Script.

PR #198, opened 2025-11-16
Add Display Model Options to Point Class Entity Definitions

RisingThumb:
This is an initial implementation, porting what I have from https://github.com/RisingThumb/func_godot_conditional_model_point_class to func_godot.

It adds a Model Descriptor resource which just contains the path, frame and skin, the scale expression is generated across all of them. It generates the meta property for `model` which is TB-specific, but I think it might be possible to expand this approach to work for other editors by doing `studio` with the same value(though I don't know as much because I don't use `studio` and I'm just going off of what the TB Docs say).

It allows you to set a default model, then a dictionary with keys which are the conditional statements, i.e. `model_id == 1` and then a corresponding Model Descriptor. This does overwrite anything set for "model" in meta properties, but _only_ if a default model or conditional models are set(which by default they are not)! This prevents it being a breaking change.

Before I do any more work on this PR cleaning up styling etc, I just want to check if this approach is a desirable approach for this problem? Because of this, I'm marking it a draft PR for now

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_point_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_point_class.gd
@@ -73,10 +73,17 @@ func model_wrangle() -> String:
 		model_switcher += " }}"
 	return model_switcher
 
+func studio_wrangle() -> String:
+	if models.size() > 1:
+		push_warning("When using a non-Trenchbroom editor, only the first Model provided is used");
+	var model := models[0]
+	return '"'+model.path+'"'
+
 
 func build_def_text(target_editor: FuncGodotFGDFile.FuncGodotTargetMapEditors = FuncGodotFGDFile.FuncGodotTargetMapEditors.TRENCHBROOM) -> String:
-	if models.size() > 0 and target_editor == FuncGodotFGDFile.FuncGodotTargetMapEditors.TRENCHBROOM:
-		self.meta_properties["model"] = self.model_wrangle();
-	# TODO: Expand for other editors
-	# could also have an option to use studio or studioprop
+	if models.size() > 0:
+		if target_editor == FuncGodotFGDFile.FuncGodotTargetMapEditors.TRENCHBROOM:
+			self.meta_properties["model"] = self.model_wrangle();
+		else:
+			self.meta_properties["studio"] = self.studio_wrangle();
 	return super(target_editor);


RhapsodyInGeek:
My initial thought was to have the model properties directly included into the Point Class resource, but giving it a bit more thought having it as a separate resource does allow reusability between objects of similar type (maybe you don't want to give enemies unique display models or maybe you want to use a shared model across all of your VFX entities, etc...).

`studio` only uses the path and has no other conditionals, so we can just autodetect with the "Target Editor" properties. Exporting from the various config resources automatically overrides the target editor to the correct one. If it's not TrenchBroom, we just fall back to only writing the default model and using the `studio` keyword.

I don't really like the recursive nature of the Model Descriptor resource though. I think Point Class can have a `display_models` Dictionary instead, with the key being the conditional and the value being the model descriptor resource. It does limit reusability a bit but I think it's a bit more clean and straightforward compromise.

Don't have time atm to get into the weeds but I do have some ideas on how to proceed with this.

RisingThumb:
I think a dictionary is less clean as you will end up with a dictionary of either arrays or dictionaries (you need to have path, frame and skin and maybe scale expressions for each) for conditional models. If done like an array it relies on the user knowing the layout of the array. If done like a dictionary it relies on the user knowing the keys for the dictionary. It would also require handling invalid Key value pairs, particularly the values.

Additionally dictionaries creates the same UX pains that class properties has where you can select many variant types(many being irrelevant) when only integer and string are relevant here and requires users to understand the model syntax format a lot more closely. It's already unwieldy for choices and spawnflags.

Finally, changing approach down the line would be a breaking change. A resource is more flexible if needs change.

I don't get the point you made about recursion? Nothing here is recursive, it just iterates over all conditional statements..?

RisingThumb:
> I'm also curious as to whether Dictionaries are guaranteed to retain their key orders, or if they're more like unordered maps. I don't really remember looking into it.

They sort of do. Quote from the [docs](https://docs.godotengine.org/en/stable/classes/class_dictionary.html)
> Dictionaries will preserve the insertion order when adding new entries. 

However the keys seem to get sorted into alphabetical order when serialised. This seems to be either a bug with Godot or an error in the documentation. You can see this issue [here](https://github.com/godotengine/godot/issues/90142). There's also UX issues in Godot with being unable to reorder the key value pairs in the editor like you can for arrays you can read the issue on [here](https://github.com/godotengine/godot-proposals/issues/13465).

I would not rely on it at the moment.

> It might be better to instead fold the condition statements into the Model Descriptor resource, and change Dictionary[String, ModelDescriptor] to Array[ModelDescriptor]. A blank condition would be treated as the default. A fail safe would be to either override or ignore any subsequently found conditionless model descriptors. Having the condition statements in the resource itself feels like less of a hack than using the Dictionary key as the condition. You also get the flexibility of being able to reorder your conditional statements.

It's between how important condition order(reordering that array) and how important not being able to represent invalid states(multiple models with duplicate conditions, probably by user error(or no condition)). I think both are important, but condition order is more important. I also don't know if when we'll see the fix for that order issue with dictionaries [seeing the lack of activity on this PR for it](https://github.com/godotengine/godot/pull/104164).

As a guard against invalid state, I could just check if a condition has been used before, and if it has, print an error.

> I'm also thinking the resource name would be better called FuncGodotFGDPointClassModelDescriptor. While it's a mouthful, it prevents class name conflict while also being descriptive enough to be obvious as to its purpose and what it belongs to and matches the naming schemes of the other resources. It also keeps it close to Point Class in the resources list.

Sounds good to me 👍

RisingThumb:
> I also think "no condition" should equate to "default". An additional check for whether it has "default" in the condition wouldn't be bad, but I think no condition should be the tutorialized default condition.

I can go ahead and add a check for whether the condition text is left as "default" if you think it's a good addition. I don't really have much of an opinion on that. Could also make the condition text default to "default" too.

I've brought the styling more in line with the rest of the func_godot codebase and renamed the model descriptors in line with your comment, and changed the approach slightly from a dictionary of conditions to ModelDescriptors to just an array of ModelDescriptors.

The next step is adding in studio/studioprop support. Doing some looking online, it looks like that only supports an individual model and can't do conditional models if I'm not mistaken? So just `studio( "pathToModel.glb" )` or `studioprop( "pathToModel.glb" )`. With this I could either use the first element in the array of model descriptors or iterate over it until I find the first default one. Either way, I personally think it's an error if the array of models has more than 1 ModelDescriptor for studio/studioprop, but before I implement it I just wanted to check I understand studio/studioprop correctly here?

Going to mark the PR as ready for review in the meantime

RhapsodyInGeek:
> I can go ahead and add a check for whether the condition text is left as "default" if you think it's a good addition. I don't really have much of an opinion on that. Could also make the condition text default to "default" too.

I'd say "don't bother" honestly. It's easy enough to leave the default condition blank and instruct people to leave it as is.

> ... it looks like that only supports an individual model and can't do conditional models if I'm not mistaken? So just `studio( "pathToModel.glb" )` or `studioprop( "pathToModel.glb" )`

That is correct. Apparently the only differences between `studio` and `studioprop` is that `studioprop` is only supported by Hammer 4.x and that it automatically determines the model's bounding box. I would suggest we ignore that one in this case. If someone wants a studioprop property bad enough for their niche use case, they can manually enter it into meta properties just fine.

We do need to support `studio` however. I think we already handle it in `build_def_text()` though, I'll have to double check.

RhapsodyInGeek:
TrenchBroom supports both `studio` and `model`, whereas no other editor supports `model`: it's a TrenchBroom exclusive. That's the reason why it only happens when `model` is used, and the main reason for the target map editor properties.

RhapsodyInGeek:
Closing as it is now implemented via 0012ef603f90587e660dda80962ba0cf88136f77

PR #209, opened 2025-12-30
Optimize Cull Interior Faces Operation with Lookup Table

RisingThumb:
Implements the optimisation discussed here https://github.com/func-godot/func_godot_plugin/pull/182#issuecomment-3694356709

The performance gains for this are significant. Where before I would not be able to blindly use this on a map that is all worldspawn, with this optimisation it is fast enough that this can be done. This would also mean we don't need to document a warning about build times while using cull interior faces.

I am also left wondering now... perhaps culling interior faces should be on by default? I've left it alone for now, but it's something to consider as I think it's a sensible default?

May need a little bit of testing still to see if any floating point issues crop up

### DIFF
--- a/addons/func_godot/src/core/geometry_generator.gd
+++ b/addons/func_godot/src/core/geometry_generator.gd
@@ -307,6 +307,17 @@ func smooth_entity_vertices(entity_index: int) -> void:
 
 #endregion
 
+func get_plane_lookup_key(plane: Plane) -> Vector4i:
+	# I don't think this constant needs to be exposed, but it avoids floating point errors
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
> perhaps culling interior faces should be on by default?

No, we should still leave it off by default as there is no necessity to having it on by default. We want to avoid potential backwards compatibility issues and on by default could cause them for some users (transparent / translucent brush entities like water and glass come to mind off the top of my head).

Looks good. I'll see about testing it this week, get it merged soon.

RisingThumb:
Transparent brushes slipped my mind, that makes sense!

PR #211, opened 2026-01-04
Fix Runtime Map Parse Failure From Carriage Returns

suhankins:
I had a bizarre issue recently where for some reason, maps I recently edited wouldn't load in exported builds anymore.

After investigating it, I believe the issue is that since the newest release of Trenchbroom, it now saves newly edited maps with `\r\n` line endings on Windows.

Godot Editor handles it and changes line endings to `\n`, but exported builds do not.

---

With this change, map parser will first try to read the map assuming it uses `\r\n`, but if it fails, it will try reading it assuming it uses `\n`.

### DIFF
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -55,9 +55,7 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 				if data.is_empty():
 					printerr("Error: Failed to open map file (" + line + ")")
 					return parse_data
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
Might be due to a recent change from #167 because Godot removed the carriage return (\r) handler from `FileAccess.get_text()` which we previously had set to `true`. I guess this is probably the knock-on effect we didn't anticipate, though I'm not positive this is necessarily the solution (it might be! I just want to give it a bit more thought).

Would also appreciate @sinewavey 's input on this.

PR #218, opened 2026-02-07
Allows FuncGodotMap to load raw map data.

greenfox1505:
This allows FuncGodotMap to load and parse raw map data directly. This is not exposed to the editor. It's only exposed in code.

This is useful for a number of reasons, but I am using it to allow compressing map files for network transfer and map storage size reduction.

An example of use (although likely you'd want something other than a `QuakeMapFile`, like a network source or compressed file source):

```gdscript
extends FuncGodotMap

# within a class that is based on FuncGodotMap
func force_rebuild():
	#local_map_file = map.resource_path
	raw_data = FileAccess.get_file_as_bytes(map.resource_path).get_string_from_utf8()
	force_build_raw = ForceBuildRaw.QUAKE_MAP
	build()
	completed_build()
```

If there is a preferred difference in how this interface should function, let me know. I tried to minimize the diff. On paper this should both VMF and MAP file types, but I didn't test VMF.

### DIFF
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -69,6 +69,28 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 		declare_step.emit("Parsing as Source VMF")
 		parse_data = _parse_vmf(map_data, map_settings, parse_data)
 	
+	# Parse the actual map.
+	return parse_map(map_data, map_settings, parse_data)
+
+
+## Parses the raw data as Quake MAP, as with `parse_map_data(`, but using raw map data instead of a file path.
+func parse_raw_map_data(raw_data: String, map_settings: FuncGodotMapSettings) -> _ParseData:
+	var map_data: PackedStringArray = raw_data.replace("\r", "").split("\n")
+	var parse_data = _ParseData.new()
+	parse_data = _parse_quake_map(map_data, map_settings, parse_data)
+	return parse_map(map_data, map_settings, parse_data)
+
+
+## Parses the raw data as Source VMF, as with `parse_map_data(`, but using raw map data instead of a file path.
+func parse_raw_vmf_data(raw_data: String, map_settings: FuncGodotMapSettings) -> _ParseData:
+	var map_data: PackedStringArray = raw_data.replace("\r", "").split("\n")
+	var parse_data = _ParseData.new()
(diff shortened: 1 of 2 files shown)


addons/func_godot/src/core/parser.gd
```diff
@@ -69,6 +69,28 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 		declare_step.emit("Parsing as Source VMF")
 		parse_data = _parse_vmf(map_data, map_settings, parse_data)
 	
+	# Parse the actual map.
+	return parse_map(map_data, map_settings, parse_data)
+
+
```
RhapsodyInGeek: Seems unnecessary to have 2 separate `parse_raw_xxx()` methods. Whether it's MAP or VMF should be auto-detected.

addons/func_godot/src/core/parser.gd
```diff
@@ -69,6 +69,28 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
...
+## Parses the raw data as Source VMF, as with `parse_map_data(`, but using raw map data instead of a file path.
+func parse_raw_vmf_data(raw_data: String, map_settings: FuncGodotMapSettings) -> _ParseData:
+	var map_data: PackedStringArray = raw_data.replace("\r", "").split("\n")
+	var parse_data = _ParseData.new()
+	parse_data = _parse_vmf(map_data, map_settings, parse_data)
+	return parse_map(map_data, map_settings, parse_data)
+
+
+## Actual map parser.
```
RhapsodyInGeek: This description doesn't sit right with me. This comment goes into the auto-doc as the description for the `parse_map()` method, and as such should state what it actually does. The `##` comments are not meant to be reminders, they're the actual documentation.

addons/func_godot/src/map/func_godot_map.gd
```diff
@@ -49,6 +56,13 @@ var _map_file_internal: String = ""
 ## Measured in Godot units, not Quake units.
 @export_range(256.0, 2048.0, 128.0) var hyperplane_size: float = 512.0
 
+## Flipping this flag will make the parser build use `raw_data` instead of a file path.
+## This is only available in code, not editor.
+var force_build_raw := ForceBuildRaw.NONE
```
RhapsodyInGeek: Feels overdesigned. Could do without this.

PR #221, opened 2026-02-19
Allows Entities to more easily reference eachother through node paths

greenfox1505:
If you add NodePath to your FGD, the system currently treats it as a string. This parses these into NodePaths to match the type in the FGD.

Additionally, it allows you to name an Entity with Godot's `%node_name` syntax. Doing so does not add `entity_` to a node, but instead uses the raw literal name as written, but trims the `%` and flips the `unique_name_in_owner` to true.

These two changes combined allow nodes to reference each other easily through familiar Godot syntax.

### Note:

I wanted to retain the `entity_` thing, but there is no way to tell at time of authoring if a reference is `entity_` or something else the scene not controlled by FuncGodot (like a UI element that might want to light up when you enter an Area or something). Dropping `entity_` for these explicate naming conditions seemed reasonable to me.

If this breaks something else that expects `entity_` name, let me know and we can try to figure out another solution.

# Examples

How this looks in Trenchbroom:

image

How it parses into Godot, notice the `@export var testTarget:NodePath` loads as reference to the other node:

image

What the FGD looks like: 

image

I don't know if this will break existing projects. If there might be another way to handle this, I'm open to changes.

### DIFF
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -56,7 +56,12 @@ func generate_solid_entity_node(node: Node, node_name: String, data: _EntityData
 	else:
 		node = Node3D.new()
 	
-	node.name = node_name
+	if node_name.begins_with("%"):
+		node_name = node_name.trim_prefix("%")
+		node.name = node_name
+		node.unique_name_in_owner = true
+	else:
+		node.name = node_name
 	node_name = node_name.trim_suffix(definition.classname).trim_suffix("_")
 	var properties: Dictionary[String, Variant] = data.properties
 	
@@ -286,7 +291,10 @@ func generate_entity_node(entity_data: _EntityData, entity_index: int) -> Node:
 	elif map_settings.entity_name_property in properties:
 		name_prop = str(properties[map_settings.entity_name_property])
 	if not name_prop.is_empty():
-		node_name = "entity_" + name_prop
+		if name_prop.begins_with("%"):
+			node_name = name_prop
+		else:
+			node_name = "entity_" + name_prop
 	
 	if entity_def is FuncGodotFGDSolidClass:
 		node = generate_solid_entity_node(node, node_name, entity_data, entity_def)
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -213,7 +213,7 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 					TYPE_STRING_NAME:
 						properties[property] = StringName(prop_string)
 					TYPE_NODE_PATH:
-						properties[property] = prop_string
+						properties[property] = NodePath(prop_string)
 					TYPE_OBJECT:
 						properties[property] = prop_string
 		
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -213,7 +213,10 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 					TYPE_STRING_NAME:
 						properties[property] = StringName(prop_string)
 					TYPE_NODE_PATH:
-						properties[property] = NodePath(prop_string)
+						if prop_string.begins_with("$") or prop_string.begins_with("%"):
+							properties[property] = NodePath(prop_string)
+						else:
+							properties[property] = prop_string
 					TYPE_OBJECT:
 						properties[property] = prop_string
 		


greenfox1505:
But I agree, I'd prefer FGDs having an array of FuncGodotClassProperty instead of dictionarys, with fixed types that more closely match the map format and descriptions bundled in. I wrote up a prototype of that with a custom editor at one point but didn't end up pushing it. It's a pain to debug custom editors in Godot.

Enums were the pain point that made me want to do that.

greenfox1505:
Yeah, I get that. It is objectively a bit of a hack.

Although using the very familiar `$` and `%` syntax, IMHO, doesn't FEEL like a "hack". It feels very natural to someone familiar with Godot. Plus a lot of the more hacky things you would need to do for this, you would also need to do for FuncGodotClassProperty, like detecting and stripping the `%` when parsing a node name and flipping the `node.unique_name_in_owner = true`. (Renaming a node in editor to `%NodeName` toggles the unique name feature, so this feels pretty normal Godot behavior).

There are project level work arounds with setters that would mean this feature wouldn't need to be part of FuncGodot itself, but those are definitely much more hacky.

```gdscript
@tool
#...
var test:String: set(v): test_path = NodePath(v)
```

greenfox1505:
I do think that at least parsing `%node_name` as a `unique_name_in_owner = true` is valuable on it's own, Even if the rest of this doesn't get in.

greenfox1505:
I know you already edited, but I just wanted to clarify:

image

You cannot tell, at FGD authoring time, if a given object name will be used as a scene-unique identifier that some other element might want to reference. The way Godot handles `%Names`, it strips the `%` and toggles that bool. If we want to use `%Names`, we'd need to do the same, regardless of the FGD.  Regardless of if we have a `FuncGodotClassProperty`. This is what I'm proposing is worth adding anyway.

`FuncGodotClassProperty` will help with treating `NodePath` and `target_destination` differently. But, IMHO, just adding a `%` or `$` to the file string also does a pretty good job of carrying that same weight with minimal added complexity.

Sidebar, not included here: I think it might also be helpful to let people explicitly name their nodes without `entity_...` so that other `NodePath`s don't get mangled. I think letting people name something with `$SomeName` is a reasonable way to do this in a non-breaking way.

RhapsodyInGeek:
Gave it another look over and some more thought. It's clean and consistent and doesn't break existing projects, and it is a Godot design pattern. Looks good to me.

greenfox1505:
I have been using this change for the past month and have noticed some issues with it. I'll create a revision when I get a chance.

RhapsodyInGeek:
What are the issues? Seems pretty straightforward what's in there.

greenfox1505:
Okay, I've had a chance to review (all in `entity_assembler.gd`):

https://github.com/greenfox1505/func_godot_plugin/commit/48898d5a22819d07560943924a57ecf100889b5b

(the below links didn't work like I hoped, you gotta scroll through and look for the blue highlighted bars on left side of the code)

- This was originally only done for `generate_solid_entity_node(` and not `generate_point_entity_node(`. So I moved some of the code to a new function that both share.
  - [That function](https://github.com/greenfox1505/func_godot_plugin/commit/48898d5a22819d07560943924a57ecf100889b5b#diff-a57972c719e3f61f7c242f1bd47e3e9954316da51149da3ebdcd129d7040e58dR43-R51)
  - [Removed from solid entity](https://github.com/greenfox1505/func_godot_plugin/commit/48898d5a22819d07560943924a57ecf100889b5b#diff-a57972c719e3f61f7c242f1bd47e3e9954316da51149da3ebdcd129d7040e58dL59-R68)
  - [Point Entity Changes](https://github.com/greenfox1505/func_godot_plugin/commit/48898d5a22819d07560943924a57ecf100889b5b#diff-a57972c719e3f61f7c242f1bd47e3e9954316da51149da3ebdcd129d7040e58dL175-R179)

- `apply_entity_properties(` needed to detect nodepath since it was still held as a string before that point (I think that's the issue here, but it's been a few weeks since I wrote it)
  - [Here](https://github.com/greenfox1505/func_godot_plugin/commit/48898d5a22819d07560943924a57ecf100889b5b#diff-a57972c719e3f61f7c242f1bd47e3e9954316da51149da3ebdcd129d7040e58dR268-R274)

- `generate_entity_node(` needed detect `$`, but keep it in the name for a later step.
  - [Here](https://github.com/greenfox1505/func_godot_plugin/commit/48898d5a22819d07560943924a57ecf100889b5b#diff-a57972c719e3f61f7c242f1bd47e3e9954316da51149da3ebdcd129d7040e58dR307-R308)
  
These changes were living in my own project repo and weren't being synced back to my fork since I wasn't sure it was going to be merged in. Living with my changes, I have found and fixed some bugs with this overall change. I can MR the fixes if you would like them.

PR #229, opened 2026-03-06
Gracefully handle multiline Quake .map properties

chow-land:
## Summary
Fixes Quake `.map` entity property parsing for multiline quoted values and escaped quotes, and prevents parser failures from cascading.

## Problem
The existing `.map` property parsing assumed key/value pairs were always on one line and relied on a fixed separator pattern.

This causes:

- incorrect parsing of multiline values,
- a crash when the closing quote appeared on its own line (`"`),
- premature value termination when escaped quotes were present,

Currently this will crash:

```txt
"message" "line one
line two
"
```

And this will parse, but line two will be dropped:

```txt
"message" "line one
line two"
```

This PR adds support to gracefully handle both of the above. It also adds support for escaped quotes, so this will parse successfully too:

```txt
"message" "line one
line \"two\"
line three"
```

### Parser
`addons/func_godot/src/core/parser.gd`

Added multiline state handling in `_parse_quake_map`:

- tracks active multiline key/value,
- appends continuation lines until an unescaped closing quote,
- reports unterminated multiline properties with line context.

Added unescaped-quote scanning helper:

- `_find_unescaped_quote(...)` ensures `\"` does not terminate the value.

Updated Quake `.map` property parsing logic:

- parses key/value boundaries by quote scanning (not fixed `" "` matching),
- accepts any amount of whitespace between key and value,
- enters multiline mode when closing quote is missing on the initial line,
- commits full multiline value including embedded newlines,
- treats malformed property lines as warnings and skips them instead of aborting the whole parse.

Added a defensive parse result check in `parse_map_data`:

- if a format-specific parser returns `null`, log an error and return an empty parse result.

## Changes unrelated to the above
  - Added `.DS_Store` to .gitignore for MacOS users

### DIFF
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -21,10 +21,10 @@ signal declare_step(step: String)
 func _find_unescaped_quote(text: String, start: int = 0) -> int:
 	var index := maxi(start, 0)
 	while index < text.length():
-		if text.unicode_at(index) == 34: # "
+		if text.unicode_at(index) == 34: # 34 = double quote (")
 			var backslash_count := 0
 			var check := index - 1
-			while check >= 0 and text.unicode_at(check) == 92: # \
+			while check >= 0 and text.unicode_at(check) == 92: # 92 = backslash (\)
 				backslash_count += 1
 				check -= 1
 			if backslash_count % 2 == 0:
@@ -49,9 +49,9 @@ func _parse_quoted_key_value_line(line: String) -> Dictionary:
 	property_data["key"] = line.substr(1, key_end - 1)
 
 	var value_open := key_end + 1
-	while value_open < line.length() and line.unicode_at(value_open) <= 32:
+	while value_open < line.length() and line.unicode_at(value_open) <= 32: # 32 = space; <= 32 skips ASCII whitespace/control chars
 		value_open += 1
-	if value_open >= line.length() or line.unicode_at(value_open) != 34:
+	if value_open >= line.length() or line.unicode_at(value_open) != 34: # 34 = double quote (")
 		return property_data
 	value_open += 1
 
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -18,6 +18,54 @@ const _ParseData	:= FuncGodotData.ParseData
 ## It is connected to [method FuncGodotUtil.print_profile_info] method if [member FuncGodotMap.build_flags] SHOW_PROFILE_INFO flag is set.
 signal declare_step(step: String)
 
+func _find_unescaped_quote(text: String, start: int = 0) -> int:
+	var index := maxi(start, 0)
+	while index < text.length():
+		if text.unicode_at(index) == 34: # "
+			var backslash_count := 0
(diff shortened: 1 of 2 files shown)


RhapsodyInGeek:
Sorry about the slow response to this PR. We've been a bit busy.

Not that I don't see value in this PR, but I'm wondering what map editors or conditions allow this to happen? I've never personally experienced this with TrenchBroom, JACK, or NetRadiant-Custom in my tests. Curious as to the situations where this issue actually appears?

chow-land:
I encountered this in the wild on a couple single player quake2 maps of mine that I've been porting to Godot. 

Basically the maps contain many NPCs with lines of dialog that are implemented via the trigger_multiple 'message' field, many of which contain long paragraphs of text. When I was composing these it was necessary to type them out in an external editor and add line breaks (both \n as well as literal line breaks, so I could plan out how they display on the screen), before pasting them back into TB. 

Also, if I'm not mistaken I believe Trenchbroom also allowed literal line breaks with shift/alt + enter in text fields previously, but you're right the latest TB version doesn't seem to allow this. 


In any case, it is a rather niche use case, but it was a fix that I needed, and it seemed like a good idea to have more bulletproof parsing anyhow.

addons/func_godot/src/core/parser.gd
```diff
@@ -18,6 +18,54 @@ const _ParseData	:= FuncGodotData.ParseData
...
+	if not line.begins_with("\""):
+		return property_data
+
+	var key_end := _find_unescaped_quote(line, 1)
+	if key_end < 0:
+		return property_data
+	property_data["key"] = line.substr(1, key_end - 1)
+
+	var value_open := key_end + 1
+	while value_open < line.length() and line.unicode_at(value_open) <= 32:
```
RhapsodyInGeek: We should comment what the magic unicode values are so we don't have to look it up when we read the code.

addons/func_godot/src/core/parser.gd
```diff
@@ -269,9 +321,34 @@ func _parse_quake_map(map_data: PackedStringArray, map_settings: FuncGodotMapSet
...
+	var multiline_property_active := false
+	var multiline_property_key := ""
+	var multiline_property_value := ""
+	var multiline_property_line := -1
 	
-	for line in map_data:
+	for line_index in map_data.size():
+		var line: String = map_data[line_index]
+		var line_number := line_index + 1
 		line = line.replace("\t", "")
```
RhapsodyInGeek: I merged a PR that breaks this one but the only difference is that this line is now
```gdscript
line = line.replace("\t", "").replace("\r", "")
```
because TrenchBroom started adding carriage returns again on Windows builds.

RhapsodyInGeek:
@chow-land whenever you get the chance to make the tweaks mentioned in my notes, I can see about getting this PR merged into the plugin.

chow-land:
@RhapsodyInGeek sure thing - unfortunately I am traveling without my personal machine right now, but I can address the comments next week.

chow-land:
@RhapsodyInGeek I added the unicode comments and rebased. Please take another look when you get a moment!

RhapsodyInGeek:
Looks good!

PR #230, opened 2026-03-06
Add map import setting to exclude sky brushes

chow-land:
## Summary
Adds a new map setting to ignore sky brushes when they match a conservative rule:
- every face on the brush uses the same texture, and
- that texture name contains `sky` in its filename. (case-insensitive)

This is intentionally conservative  as to avoid filtering mixed-texture brushes, i.e. a brush that only has some faces with a sky texture.

Personally I would find this quite useful when I'm importing maps that were built for the base quake games, as I always end up going through and deleting all of the sky brushes manually in Trenchbroom before I import to Godot, since they usually serve no purpose.

## Changes
- Added `exclude_sky_brushes` to map settings:
  - `addons/func_godot/src/map/func_godot_map_settings.gd`
- Added brush filter helper in parser:
  - `addons/func_godot/src/core/parser.gd`
- Applied the filter at brush commit points for both parsers:
  - Quake MAP parser (`_parse_quake_map`)
  - VMF parser (`_parse_vmf`)

'Exclude Sky Brushes' == OFF
Screenshot 2026-03-05 at 11 25 08 PM

'Exclude Sky Brushes' == ON
Screenshot 2026-03-05 at 11 25 15 PM

### DIFF
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -18,6 +18,19 @@ const _ParseData	:= FuncGodotData.ParseData
 ## It is connected to [method FuncGodotUtil.print_profile_info] method if [member FuncGodotMap.build_flags] SHOW_PROFILE_INFO flag is set.
 signal declare_step(step: String)
 
+func _is_excluded_sky_brush(brush: _BrushData, map_settings: FuncGodotMapSettings) -> bool:
+	if not map_settings or not map_settings.exclude_sky_brushes:
+		return false
+	if not brush or brush.faces.is_empty():
+		return false
+	var texture_name: String = brush.faces[0].texture.to_lower()
+	if not texture_name.contains("sky"):
+		return false
+	for face in brush.faces:
+		if face.texture.to_lower() != texture_name:
+			return false
+	return true
+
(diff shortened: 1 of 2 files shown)


RhapsodyInGeek:
We appreciate all the recent PRs, but this particular one will have to be rejected.

One of the things we try to avoid doing is "designing the game for the developer". This would be one of those cases, where "auto-handling sky textures" is overreaching even as an optional setting. It's up to the dev to determine how the sky is built and processed, not the tool.

We also try to avoid duplicating functionality, and this essentially duplicates the skip / clip features. If you're looking to import Quake (or other maps) that use the sky texture and want to remove them, an easy enough solution is to use the map editor's replace texture functions to replace them all sky textures with your skip or clip texture. This will ensure the mesh faces do not get built by func_godot.

We try to be cautious about adding new properties and settings, as func_godot is already pretty filled to the brim with them (MapSettings is already quite bloated). Instead we tend to gravitate towards expanding the functionality of current settings or educating users on the ways certain goals can already be accomplished.

chow-land:
Ah ok that's fair, I'll just use this locally then

RisingThumb:
You can also do this without changing the addon by including texture name metadata on your solid class entities, see if it has a sky texture, and if so queue it to be freed while building. This way you can also expand it to more texture names too

PR #233, opened 2026-03-15
fix maps containing \r not reading correctly from exported builds

ottworks:
building maps at runtime from exported builds wasn't working for me because the map files had carriage returns in them, preventing the keyvalue parser from correctly stripping trailing double quotes. this pr fixes that by remiving carriage returns.


RhapsodyInGeek:
This PR is unacceptable. We need a description of what the issue you're trying to solve is and what your fix actually entails, and why you believe it's needed. Your title needs to be more succinct; continuing your title with an ellipsis into the post is also unacceptable.

RhapsodyInGeek:
Looks good.

PR #235, opened 2026-03-20
Include subdirectories when searching for textures

UrbanNuke:
As title says

### DIFF
--- a/addons/func_godot/src/util/func_godot_util.gd
+++ b/addons/func_godot/src/util/func_godot_util.gd
@@ -138,7 +138,7 @@ static func load_texture(texture_name: String, wad_resources: Array[QuakeWadFile
 	
 	return load(default_texture_path)
 	
-## searching for subdirectories and multiple file extensions.
+## Searching for subdirectories and multiple file extensions.
 static func load_texture_recursive(texture_name: String, base_dir: String, extensions: Array) -> Texture2D:
 	var dirs: PackedStringArray = DirAccess.get_directories_at(base_dir)
 	if dirs.is_empty():
--- a/addons/func_godot/src/util/func_godot_util.gd
+++ b/addons/func_godot/src/util/func_godot_util.gd
@@ -126,14 +126,10 @@ const _pbr_features: PackedInt32Array = [
 ## Searches for a Texture2D within the base texture directory or the WAD files added to map settings. 
 ## If not found, a default texture is returned.
 static func load_texture(texture_name: String, wad_resources: Array[QuakeWadFile], map_settings: FuncGodotMapSettings) -> Texture2D:
+	var texture: Texture2D = load_texture_recursive(texture_name, map_settings.base_texture_dir, map_settings.texture_file_extensions)
 	for texture_file_extension in map_settings.texture_file_extensions:
-		var texture_path: String = map_settings.base_texture_dir.path_join(texture_name + "." + texture_file_extension)
-		if ResourceLoader.exists(texture_path):
-			var texture_file = load(texture_path)
-			if texture_file is Texture2D:
-				return texture_file
-			else:
-				printerr("Error: Texture load failed! (%s) not a valid Texture2D resource", texture_path)
+		if texture != null:
+			return texture
 	
 	var texture_name_lower: String = texture_name.to_lower()
 	for wad in wad_resources:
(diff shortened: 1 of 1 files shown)


RhapsodyInGeek:
Not an acceptable PR. Need to know what the problem was, why you think this change is needed, and how this change works at a minimum.

This one particularly so. We don't search subdirectories by design, and we would need an overriding reason to change that.

UrbanNuke:
The problem is that when using “Build Map” in FuncGodotMap, the textures used on the map are only searched for in the top-level directory specified in FuncGodotMapSettings -> “Base Texture Dir,” and do not look in subdirectories. It can be very inconvenient to keep all textures in a single directory.

This has now been addressed, and all nested directories are now taken into account

RisingThumb:
I haven't looked closely at the code, but as a small thing that might be missed, if you set a separate directory for the materials and save generated materials, are they saved in matching subdirectories for that materials folder?

UrbanNuke:
No, materials are saved without taking into account the nested texture directories

RisingThumb:
So if I understand correctly, say you have a texture in `res://textures/stone/stone_walls/stone_wall_cracked.png` this wouldn't map to a material in `res://materials/stone/stone_walls/stone_wall_cracked.tres`? What does it map to then?

RhapsodyInGeek:
> The problem is that when using “Build Map” in FuncGodotMap, the textures used on the map are only searched for in the top-level directory specified in FuncGodotMapSettings -> “Base Texture Dir,” and do not look in subdirectories. It can be very inconvenient to keep all textures in a single directory.
> 
> This has now been addressed, and all nested directories are now taken into account

I have never experienced this issue. I am not sure why you are.

If I have a texture defined in the map file as `env/grass` and my map settings has `res://textures/` set as my textures folder, FuncGodot looks for `res://textures/env/grass.png`.

If I have my materials folder set as `res://materials/` it will save the generated material as `res://materials/env/grass.tres`.

This is something that has been working on the latest code base as of earlier this week.

In other projects I have been able to nest textures deeper.

The current path search it performs is `map_settings.textures_dir + map_texture_path + file_extensions`. This has been working as intended.

I'm not sure why this wouldn't be working for you, or if you mean something else isn't working how you expect it to?

UrbanNuke:
> This is something that has been working on the latest code base as of earlier this week.

I used last release 2025.12, and I expected it to work, but it didn't.

I just have map settings config with
BaseTextureDir: res://tb/textures
BaseMaterialDir: res//tb/materials

In base texture dir I have no textures on top level, but have two folders "test" and "special".
Everything works fine with my changes, if I remove them - "Build Map" can't find textures and all my objects are using default material.

I've tried replace addons/func_godot/src/util/func_godot_util.gd by last version from the main branch (because it was updated recently) and got the same result.


> The current path search it performs is `map_settings.textures_dir + map_texture_path + file_extensions`. This has been working as intended.

That's not quite right. In the current implementation, `texture_path` is defined as `map_settings.base_texture_dir.path_join(texture_name + “.” + texture_file_extension)`
addons/func_godot/src/util/func_godot_util.gd:130

There is only texture name, or might be I'm cooking Trenchbroom wrong, but it looks fine inside Trenchbroom.
Potentially Trenchbroom should name textures with relative path


> So if I understand correctly, say you have a texture in `res://textures/stone/stone_walls/stone_wall_cracked.png` this wouldn't map to a material in `res://materials/stone/stone_walls/stone_wall_cracked.tres`? What does it map to then?

Yes, I got materials in base materials dir.
In my case textures from res://tb/textures/test/sand.png placed into res://tb/materials/sand.tres

RhapsodyInGeek:
> That's not quite right. In the current implementation, texture_path is defined as map_settings.base_texture_dir.path_join(texture_name + “.” + texture_file_extension)
addons/func_godot/src/util/func_godot_util.gd:130

That's exactly what I described. `texture_name` is a bit of a misnomer in the sense that it's really a path to the texture relative to the base textures directory, minus the file extension.

The selected texture in the TrenchBroom screenshot shows `env/grass0` is the currently applied texture.
image

When viewed in a text editor, we can see the value put in the map file:
```
// brush 5
{
( 0 -64 -16 ) ( 0 -63 -16 ) ( 0 -64 -15 ) skip [ 0 1 0 0 ] [ 0 0 -1 0 ] 0 1 1
( 64 0 16 ) ( 64 0 17 ) ( 65 0 16 ) skip [ 1 0 0 0 ] [ 0 0 -1 0 ] 0 1 1
( -64 -64 -16 ) ( -63 -64 -16 ) ( -64 -63 -16 ) skip [ 1 0 0 0 ] [ 0 -1 0 0 ] 0 1 1
( 64 64 0 ) ( 64 65 0 ) ( 65 64 0 ) env/grass0 [ 1 0 0 0 ] [ 0 -1 0 0 ] 0 1 1
( 64 640 16 ) ( 65 640 16 ) ( 64 640 17 ) skip [ 1 0 0 0 ] [ 0 0 -1 0 ] 0 1 1
( 640 64 16 ) ( 640 64 17 ) ( 640 65 16 ) skip [ 0 1 0 0 ] [ 0 0 -1 0 ] 0 1 1
}
```
`env/grass0` is the value that gets passed to `texture_name` in the current implementation.

It was autogenerated along with every other material you see in the screenshot below, except for the sky and water materials. The vine materials were autogenerated and modified afterward, but not moved.
image

So everything is already working the way it should with subdirectories. I do not know why you're having issues, besides you possibly having an incorrect Map Settings resource applied to your FuncGodotMap node, or if your TrenchBroom configuration is set up in a way that causes your textures to have incorrect paths in the output map file.

UrbanNuke:
Ah, I guess we found my problem. In Trenchbroom I have texture name without relative path
image
Do you know how to solve it?

UrbanNuke:
Okay, sorry for taking your time, but I just haven't assigned "Map Editor Game Path" property in FuncGodotLocalConfig.
That's it

PR #237, opened 2026-04-02
Reduce precision of coplanar detection in cull_interior_faces

ottworks:
My map was getting lots of interior faces for unknown reasons. This change fixes that and makes cull_interior_faces more reliable.


RhapsodyInGeek:
Looks good!

PR #238, opened 2026-04-10
Fixed aabb size and position miscalculations

StepingForward:
My issue was that the previous AABB calculations didn't take into account the global position and scale of the node in the scene. 
These changes fix that. Now the entire node3d's transform is taken into account on generation. 
| Before:  | After:  |
| ------------- | ------------- |
| image | image |


RhapsodyInGeek:
Looks good!

PR #240, opened 2026-04-12
Moved property type definition for "target_source"

Adrihfu:
Related to this issue #215 
where Godot saves the `Object` Variant Type has an `null` in file when empty.

Changing the type fixes it.

It was moved from `TYPE_OBJECT` to `TYPE_STRING_NAME`, which was not been used and fits the use of `target_source`.

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -215,9 +215,9 @@ func build_def_text(target_editor: FuncGodotFGDFile.FuncGodotTargetMapEditors = 
 						prop_type = "decal"
 					elif value is AudioStream:
 						prop_type = "sound"
-				else:
-					prop_type = "target_source"
-					prop_val = "\"\""
+			TYPE_STRING_NAME:
+				prop_type = "target_source"
+				prop_val = "\"\""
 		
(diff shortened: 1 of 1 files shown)


suhankins:
`TYPE_STRING_NAME` is used as `StringName`. I.e. if I have an object with property `action` that takes `StringName`, I'd need to set `action` in my FGD config to be `StringName`

Adrihfu:
> `TYPE_STRING_NAME` is used as `StringName`.

Inside the `func_godot_fgd_entity_class.gd` script `StringName` was not been check.


> I.e. if I have an object with property `action` that takes `StringName`, I'd need to set `action` in my FGD config to be `StringName`

By default `StringName` is not been added to the `FGD` because is not been check in the `func_godot_fgd_entity_class.gd`.
temp
temp2

You mean that a class or script that you are using recieves a `StringName` variable named `action`, right?

suhankins:
> You mean that a class or script that you are using recieves a `StringName` variable named `action`, right?

I was wrong about how it would work, my bad.

RhapsodyInGeek:
TYPE_STRING_NAME is actually used by the Parser, as an alternate for the string type. The main reason is that even though the map file reads it as any standard string, when imported into Godot it will convert the map string to a Godot StringName. This PR is incompatible unfortunately.

PR #243, opened 2026-04-22
Adds type check on `load(` in the map resource importer

greenfox1505:
I ran into this issue on my project since I use a different map importer. This lets you go back to FuncGodot's QuakeMapFile.


RhapsodyInGeek:
Looks good!

PR #244, opened 2026-05-04
Fix scene needing a manual save after map build/clear in Godot 4.5 and higher

### LINKED ISSUE #242: Scene needs manually saving after building in Godot 4.5 and higher
The issue seems to be from `EditorInterface.mark_scene_as_unsaved()` needing to be set in func_godot.
I opened a topic about this for TBLoader (based on the original Qodot) last year: https://github.com/codecat/godot-tbloader/issues/115#issuecomment-4276479784

It was solved here: https://github.com/codecat/godot-tbloader/issues/115#issuecomment-4276479784

```swift
diff --git a/addons/tbloader/src/plugin.gd b/addons/tbloader/src/plugin.gd
index 9844925..38cca45 100644
--- a/addons/tbloader/src/plugin.gd
+++ b/addons/tbloader/src/plugin.gd
@@ -37,3 +37,4 @@ func create_map_control() -> Control:
 func build_meshes():
        var loader = editing_loader.get_ref()
        loader.build_meshes()
+       EditorInterface.mark_scene_as_unsaved()
```

While func_godot is different to Qodot, it seems to have the same issue with needing to save the scene manually after building the map. 

I also opened the issue in Godot bug reports, but it's been confirmed it's an issue with the plugin: https://github.com/codecat/godot-tbloader/issues/115?issue=godotengine%7Cgodot%7C104688#issuecomment-4276479784

CarpetBakery:
Fixes #242. Added `EditorInterface.mark_scene_as_unsaved()` to `clear_children`, which is called at the beginning of a map build.

### DIFF
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -61,6 +61,7 @@ func clear_children() -> void:
 	for child in get_children():
 		remove_child(child)
 		child.queue_free()
+	EditorInterface.mark_scene_as_unsaved()
 
 ## Checks if a [QuakeMapFile] for the build process is provided and can be found.
 func verify() -> Error:


RhapsodyInGeek:
Looks good!

PR #248, opened 2026-06-11
Add checks for empty and duplicate FGD class names

### LINKED ISSUE #247: FGD containing empty or duplicate classname may be exported without error
Currently, it is possible to export a game config / FGD setup containing either an empty classname field (which will thrown an error in TrenchBroom and Radiant) or a duplicate of a previously existing (throws a warning in TrenchBroom, silently fails in NRC and uses the first-exported def).

sinewavey:
Resolves #247

This PR adjusts a few things related to entity FGD exporting when either an empty or duplicate classname has been found.

`FuncGodotFGDFile::do_export_file` now returns an `Error` type to indicate export failures, and will skip opening and writing to a file should an error be encountered. Both `TrenchBroomGameConfig` and `NetRadiantCustomGamePackConfig` now check for this error and report the failure.

Associated error handling messages in `FuncGodotFGDFile` have been adjusted to contain more specific information regarding where errors occurred to allow users to more easily pinpoint and resolve.

Unrelated to the FGD export related changes, some instances of `push_error` were adjusted to `printerr`, as the former will include an unnecessary link to Godot's inbuilt variant source code.

On the whole, this should likely be adjusted throughout the entire plugin, though it is not nearly as pressing.

### DIFF
--- a/addons/func_godot/src/fgd/func_godot_fgd_file.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_file.gd
@@ -22,12 +22,12 @@ enum FuncGodotTargetMapEditors {
 func export_button() -> void:
 	do_export_file(target_map_editor)
 
-func do_export_file(target_editor: FuncGodotTargetMapEditors = FuncGodotTargetMapEditors.TRENCHBROOM, fgd_output_folder: String = "") -> void:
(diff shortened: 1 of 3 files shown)


RhapsodyInGeek:
Looks good!

PR #251, opened 2026-06-30
Specify Resource types on import

Visssarion:
Changes generic `Resource` string to specific string of imported resource.

This allows for `EditorFileSystem.get_file_type()` string comparison to detect a specific resource type.


RhapsodyInGeek:
Does this have any impact on exported projects?

Visssarion:
I don't think it does, as I understand that string is just a hint for Editor.

I checked this version on my game (exported build) and everything worked the same.

RhapsodyInGeek:
Looks good to me, sorry for the delay!

PR #254, opened 2026-07-03
make is_smooth_shaded type safe

Catboy-Slim:
properties.get(smoothing_property, 0) returns String "0", which causes an error for me


sinewavey:
Thanks for the contribution, and apologies for the delay in review. Looks good to me, though before merge, is there a reason to choose int(result) > 0 over bool(result)?

Catboy-Slim:
simply because Trenchboom uses ints for bools

RhapsodyInGeek:
> Thanks for the contribution, and apologies for the delay in review. Looks good to me, though before merge, is there a reason to choose int(result) > 0 over bool(result)?

I haven't tested yet, but if I recall correctly `bool("0")` will return `True` because technically the String isn't empty.

That said, I'm also not convinced that this PR is necessarily solving the real problem either, as I don't remember if `int("1")` actually returns `1` or returns `0`?

Catboy-Slim:
String to int does parse correctly (within reason, it simply removes all non-[0-9.] first)

image

RhapsodyInGeek:
Looks good!

PR #259, opened 2026-07-26
FGD: if Object is unknown Resource, default to target_source

Catboy-Slim:
When I set my type to Object, it's not saved and reverted when i re-open godot. I  set it to Resource instead, but this produces a syntax error due to the res:// path. This change makes it a target_source if the Resource type is unknown, because it would have produced a syntax error anyways.


RhapsodyInGeek:
We're looking into an alternative solution for this.

Issue #256, opened 2026-07-12
palette option for the TrenchBroom gameconfig should be optional


RisingThumb:
See this issue on the [TrenchBroom issue tracker for more context](https://github.com/TrenchBroom/TrenchBroom/issues/5335#issue-4867954775)

Currently Func_Godot ensures that Palette is included and the path for it is "textures/palette.lmp". If you do not have a palette.lmp located there, you will have an error in the log for TrenchBroom: `Could not reload material collections: textures/palette.lmp not found`. In addition textures are not loaded and rendered within TrenchBroom when this is missing. Seeing the above issue, the palette is optional for the game config and when it isn't set, TrenchBroom handles it fine.

This issue occurs using 2026.1 of TrenchBroom. Using 2025.2 this issue also goes away. Will also raise upstream just to check if this is intended.

Current work around is to include the palette.lmp file in your textures folder even if it will never be used by your game. Ideally, this is fixed by making the palette in the game configuration an optional entry. I would also lean to saying this should default to being off by default as I believe most func_godot users are not making games that depend on palette.lmp being defined.

RhapsodyInGeek:
We're still undecided on a stance regarding official continued support for future versions of TrenchBroom beyond 2025.2 due to the maintainer's switch to AI / LLM coding and doubling down on supporting it. I personally won't be a part of any direct support for TrenchBroom from here on out. Anyone who wishes to continue further TrenchBroom support will more than likely have to do it on their own forks. Fortunately, 2025 still works perfectly fine.

We're currently waiting for updates regarding [the NextBroom fork](https://codeberg.org/NextBroomTeam), which we'll have official support for after they have their first proper release. There still seems to be activity by several contributors, so hopefully we'll see it pick up steam once they get a few releases under their belt.

As far as continuing support for Godot past 4.6, while we haven't discussed it as a team I personally feel a lot more comfortable continuing with it after Godot Foundation updated their policy and stance on AI usage for contributions.

While we can't prevent all of our dependencies from having some type of GenAI contribution at some leg in their pipeline or development, where we *can* control it, we should. In regards to TrenchBroom, 2025.2 still works just fine, and hopefully we'll have an alternative soon enough that we won't even need to concern ourselves over it anymore.

RisingThumb:
I have been thinking a little bit about this. How does this sound as an approach?

[...]

PR #260, opened 2026-07-30
Default `TrenchBroomGameConfig.palette_path` to empty string.

greenfox1505:
If this file does not exist, it causes the latest TrenchBroom to fail to load materials.

TrenchBroom does not consider this a bug and will not fix it: https://github.com/TrenchBroom/TrenchBroom/issues/5335

Making this string default to empty will make prevent loading issues.


RisingThumb:
This looks good to me and it fixes #256

PR #261, opened 2026-08-06
Better method of specifying a target source property type for FGD entities

redxdev:
This is an alternate take on fixing #215 as opposed to using `StringName` like #240 and I believe sets a better standard for how to handle "extended" property types going forward.

The idea is that to make a target_source property you make a new inline resource of type `FuncGodotTargetSourceProperty` instead of an `Object` (since that won't save properly and gets reset to null on reload of the entity definition resource).

image

A benefit of this approach for future property type support is that additional options could be made available in the resource, such as a proper default value instead of it always being empty.

An alternate approach would be to have a more generic extended property data resource type for specifying custom FGD types like so:

```gdscript
class_name FuncGodotPropertyData
extends Resource

@export var prop_type: String
@export var prop_val: String
```

but that requires a bit more work.

### DIFF
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -297,9 +297,12 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 						properties[property] = prop_default[prop_default.keys().front()]
 					else:
 						properties[property] = 0
-				# Materials, Shaders, and Sounds
+				# Materials, Shaders, Sounds, and special property types
 				elif prop_default is Resource:
-					properties[property] = prop_default.resource_path
+					if prop_default is FuncGodotTargetSourceProperty:
+						properties[property] = ""
+					else:
+						properties[property] = prop_default.resource_path
 				# Target Destination and Target Source
 				elif prop_default is NodePath or prop_default is Object or prop_default == null:
 					properties[property] = ""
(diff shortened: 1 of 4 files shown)


RisingThumb:
@redxdev could this theoretically be expanded to also have a resource covering the choices class property since it's definition is clunky and requires setting stuff in the class property description.

Secondly, can this be adopted while maintaining backwards compatibility? I believe it can be right?

Edit: if it's yes to both, I'm strongly in favour of this change

redxdev:
@RisingThumb this change does not itself enable that, but there's no reason the choices type (and other clunky types) couldn't follow this exact pattern. I agree that replacing the existing setup for choices with a new resource type would make them much easier to work with.

sinewavey:
Yeah, I do think choices/flags could be handled a bit more gracefully.

One thing to note: The way NRC handles flag properties (requiring a name of spawnflags, and having overlaps can cause problems) might make the implementation a bit more complex. If we can work around that and ensure it still works with NRC, I'm in favor of making these property types easier to set up in Godot.

redxdev:
Actually, choices being a resource like in this PR could be quite useful - often choices represent an enum that might be used in a few places, and instead of having to define the setup for that multiple times across different entity types you'd be able to point them all at the same choices resource.

Unfortunately for the purpose of _this_ PR there are no real benefits to `FuncGodotTargetSourceProperty` being a resource because there's nothing to configure on it and nothing that could be shared across entity types, but alas it's what we have to work with as far as custom types in the inspector...

redxdev:
I've made a new PR for implementing choices with the same technique as this PR ^

Flags should also be doable with the same technique, I just haven't gotten around to it.

PR #262, opened 2026-08-06
Handle Map files as PackedScenes

erodozer:
- Adopts code from https://github.com/RisingThumb/func_godot_scene_import with adjustments
- Changes manual generation of maps through the editor interface to use Godot's standardized Scene Import processing
- Decouples generation logic from SceneTree for full headless factory style generation
- Generation code moved entirely out of FuncGodotMap into ~~MapSceneImportPlugin~~ FuncGodotMapLoader
- ~~FuncGodotMap retained for type capability purposes, though lacks any functionality. It is set as the default script type for Maps by the import plugin~~ FuncGodotMap functionality acts as a facade to Map Loader, deprecation notice recommends Importing maps as Packed Scenes instead of using destructive loading.

Discussion: https://github.com/orgs/func-godot/discussions/201

~~Unlike proposed in the discussion, this PR does introduce the breaking change of moving generation logic out of FuncGodotMap.~~  This PR is to better follow Godot best practices around scene resources.


RisingThumb:
> Unlike proposed in the discussion, this PR does introduce the breaking change of moving generation logic out of FuncGodotMap. This is purposely done to better adopt Godot standardized functionality and best practices around scene resources.

From least bad to worst:

- This breaks projects extending FuncGodotMap. As an [example](https://github.com/milkru/kajmak)
- This breaks peoples current projects using the node system, requiring a lot of time and likely error-prone work updating their scenes to use this.
- This removes runtime map generation as the Godot Importer is not available at runtime.

I'm ok with having .map files have an option to be imported as scenes, but following some standards better doesn't outweigh the cost of the above 3 issues.

erodozer:
> From least bad to worst:
> 
>     * This breaks projects extending FuncGodotMap. As an [example](https://github.com/milkru/kajmak)

I'm less worried about plugins like this which extend FuncGodotMap, as the API appears more for internal use versus the purpose of leveraging Maps as assets.  That said, refactoring the plugin should be straightforward since functionality largely hasn't been rewritten, it's mostly just being shuffled around.

>     * This breaks peoples current projects using the node system, requiring a lot of time and likely error-prone work updating their scenes to use this.
>     * This removes runtime map generation as the Godot Importer is not available at runtime.

Logic has now been moved outside of the Importer plugin so it's accessible at runtime.  Functionality has been restored to FuncGodotMap to avoid any project breakage.  Both FuncGodotMap and the Importer are implemented as facades to the Loader.

Deprecation notice on FuncGodotMap added to encourage usage of calling the Loader at runtime through scripts instead of through the FuncGodotMap node, as well as to use the Importer for the sake of treating them as PackedScenes.

RhapsodyInGeek:
I'm not convinced this is the right direction for the plugin. RisingThumb's earlier proposal was more appealing because of the option being provided without breaking projects or pushing anyone to change their processes, and it was provided as an alternative option rather than the intended default.

I'll leave this open for now, as open discussion is always good to have on this project.

RhapsodyInGeek:
Closing due to inactive discussion and feature not planned.

PR #264, opened 2026-08-10
Add configuration to inject func_godot_properties into entity metadata 

erodozer:
Support workflows that avoid custom scripts per entity type by injecting entity properties into object meta fields as a ~~fallback~~ configurable behavior.

My usecase is that architectually most of my entities are data containers, with behavior hoisted elsewhere.  It's largely unnecessary for me to make godot scripts for each one, so instead I opt to keep things as plain node types.  Still, I need to get the entity properties somehow at runtime, and it's more flexible to just access it through generic storage available on all godot object types.

### DIFF
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -278,6 +278,9 @@ func apply_entity_properties(node: Node, data: _EntityData) -> void:
 	if "func_godot_properties" in node:
 		node.func_godot_properties = properties
 	
+	if data.definition.add_properties_to_metadata:
+		node.set_meta("func_godot_properties", properties)
+
 	if node.has_method("_func_godot_apply_properties"):
 		node.call("_func_godot_apply_properties", properties)
 	
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -61,6 +61,9 @@ var prefix: String = ""
 ## This can be used to programmatically generate class properties when you need it
 @export var entity_extension_script: Script
 
+## Attaches entity properties from the entity onto the output node's metadata as `"func_godot_properties"`
+@export var add_properties_to_metadata : bool = false
+
 ## Parses the definition and outputs it into the FGD format.
 func build_def_text(target_editor: FuncGodotFGDFile.FuncGodotTargetMapEditors = FuncGodotFGDFile.FuncGodotTargetMapEditors.TRENCHBROOM) -> String:
 	# Class prefix


RisingThumb:
The entity class should have a toggle for whether you want the properties attached to the metadata. As is, this inflates what each node in a generated map stores which for me isn't desirable behaviour.

erodozer:
> The entity class should have a toggle for whether you want the properties attached to the metadata. As is, this inflates what each node in a generated map stores which for me isn't desirable behaviour.

Added a config toggle on the entity class to control this fallback behavior

addons/func_godot/src/core/entity_assembler.gd
```diff
@@ -277,7 +277,9 @@ func apply_entity_properties(node: Node, data: _EntityData) -> void:
 	
 	if "func_godot_properties" in node:
 		node.func_godot_properties = properties
-	
+	elif data.definition.properties_as_meta:
```
redxdev: Should this be an `if` instead of an `elif`? If you've opted into the config option then it should probably apply regardless of whether the node has a `func_godot_properties` property.

erodozer: My intention was for this to just be a fallback behavior because if you have a script attached to the entity with the field then you are likely designing around know it has that.

However, I can understand wanting to make the data always available in meta if checked.  That would allow for simpler post-processing across entities that utilized shared properties without needing to duplicate this conditional logic.

redxdev: Yeah, I understand the original intention and it makes sense, but imo now that you've added a config option it should be treated less like a fallback and more like "the user has specifically requested this behavior".

sinewavey: I would agree with redxdev - while I do like the option to use metadata (and the concept of this PR).

erodozer: I've updated behavior to always add the meta when configured instead of only as fallback, even when the field on the node exists

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
```diff
@@ -61,6 +61,9 @@ var prefix: String = ""
 ## This can be used to programmatically generate class properties when you need it
 @export var entity_extension_script: Script
 
+## Attaches entity properties from the entity onto the output node's metadata as `"func_godot_properties"`
+@export var properties_as_meta : bool = false
```
RhapsodyInGeek: Should be named `add_properties_to_metadata`, to follow the more descriptive naming conventions throughout the plugin. FuncGodot classes should have extremely explicit and obvious names.

erodozer: done https://github.com/func-godot/func_godot_plugin/pull/264/commits/4aa8df090ffeb71c2affd53bb3e464cf46089073

sinewavey:
looks good to me; any other thoughts before merge @RhapsodyInGeek ?

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
```diff
@@ -61,6 +61,9 @@ var prefix: String = ""
 ## This can be used to programmatically generate class properties when you need it
 @export var entity_extension_script: Script
 
+## Attaches entity properties from the entity onto the output node's metadata as `"func_godot_properties"`
+@export var add_properties_to_meta : bool = false
```
RhapsodyInGeek: `add_properties_to_metadata` please. We need to be specific and pedantic about this because once it's merged, the property name doesn't change, and in Godot it's referred to as metadata specifically.

erodozer: renamed in latest push

RhapsodyInGeek:
Sorry to keep you waiting! Looks good to me!

PR #265, opened 2026-08-10
Use a signal to trigger _func_godot_build_complete instead of using a deferred call

redxdev:
This fixes some hidden issues with using func_godot as part of importing a scene (like with https://github.com/RisingThumb/func_godot_scene_import). If you try to run godot in import mode on the command-line (via `--import --headless`) the engine can finish importing a scene and quit before any deferred calls have a chance to run, thus resulting in `_func_godot_build_complete` never getting called on any entities.

Relying on deferred calls during a map build seems a bit iffy to me anyway, and means that `FuncGodotMap.build()` returns before everything has actually completed, which can lead to some weird bugs in user code.

There are a few ways I could have gone about this, but a signal that gets hooked up automatically seemed like a reasonable solution to me.

### DIFF
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -19,6 +19,9 @@ var build_flags: int = 0
 ## It is connected to [method FuncGodotUtil.print_profile_info] method if [member FuncGodotMap.build_flags] SHOW_PROFILE_INFO flag is set.
 signal declare_step(step: String)
 
+## Emitted when the map build is complete. [code]_func_godot_build_complete()[/code] is automatically connected to this for generated entities.
+signal build_complete
+
 func _init(settings: FuncGodotMapSettings) -> void:
 	map_settings = settings
 
@@ -250,7 +253,7 @@ func generate_point_entity_node(node: Node, node_name: String, properties: Dicti
 
 ## Converts the [String] values of the entity data's [code]properties[/code] [Dictionary] to various [Variant] formats 
 ## based upon the [FuncGodotFGDEntity]'s class properties, then attempts to send those properties to a [code]func_godot_properties[/code] [Dictionary] 
-## and an [code]_func_godot_apply_properties(properties: Dictionary)[/code] method on the node. A deferred call to [code]_func_godot_build_complete()[/code] is also made.
+## and an [code]_func_godot_apply_properties(properties: Dictionary)[/code] method on the node. A call to [code]_func_godot_build_complete()[/code] will be made once the rest of the map build completes.
 func apply_entity_properties(node: Node, data: _EntityData) -> void:
 	var properties: Dictionary[String, Variant] = data.properties
 	
@@ -282,7 +285,7 @@ func apply_entity_properties(node: Node, data: _EntityData) -> void:
 		node.call("_func_godot_apply_properties", properties)
 	
 	if node.has_method("_func_godot_build_complete"):
-		node.call_deferred("_func_godot_build_complete")
+		build_complete.connect(node.get("_func_godot_build_complete"), CONNECT_ONE_SHOT)
 
 ## Generate a [Node] from [FuncGodotData.EntityData]. The returned node value can be [code]null[/code], 
 ## in the case of [FuncGodotFGDSolidClass] entities with no [FuncGodotData.BrushData] entries.
--- a/addons/func_godot/src/map/func_godot_map.gd
+++ b/addons/func_godot/src/map/func_godot_map.gd
@@ -142,6 +142,7 @@ func build() -> void:
 		print("\nENTITY ASSEMBLER")
 		assembler.declare_step.connect(FuncGodotUtil.print_profile_info.bind(assembler._SIGNATURE))
 	assembler.build(self, entities, groups)
+	assembler.build_complete.emit()
 	
 	time_elapsed = Time.get_ticks_msec() - time_elapsed
 


RisingThumb:
Can you verify order of operations across a number of different entities?

Order should be every entity's _func_godot_apply_properties() gets called then after they're all done, each entity's _func_godot_build_complete() is then called afterwards. I believe the position of the signal being emitted means it should do that, but I'm not 100%.

Additionally these signals should be disconnected afterwards. If you reparent the children to outside of the FuncGodotMap then call build() in editor or at runtime, this will result in it being called a 2nd time which shouldn't happen.

redxdev:
@RisingThumb the signal is emit at the end of the build, so there's no way it can be emit before an entity's `_func_godot_apply_properties` has been called.

> Additionally these signals should be disconnected afterwards.

I believe `CONNECT_ONE_SHOT` being passed should already do that.

addons/func_godot/src/core/entity_assembler.gd
```diff
@@ -19,6 +19,9 @@ var build_flags: int = 0
 ## It is connected to [method FuncGodotUtil.print_profile_info] method if [member FuncGodotMap.build_flags] SHOW_PROFILE_INFO flag is set.
 signal declare_step(step: String)
 
+## Emitted when the map build is complete. [code]_func_godot_build_complete()[/code] is automatically connected to this for generated entities.
+signal build_complete
```
RhapsodyInGeek, line 23: Signal should be `entity_assembly_complete`, as there is already a `build_complete` signal on the  `FuncGodotMap` node.

RhapsodyInGeek:
Main problem with this that I can see is tree modification happening when generating new nodes within other entities. The main reason for deferring the call to `_func_godot_build_complete()` instead of calling it after all the entity's have run their `_func_godot_apply_properties()` functions is because the SceneTree might be locked from adding new children or modifying existing children during the first phase.

I have a lot of entities that require this that I don't think this change will work with, because I do _a lot_ of SceneTree modification during the build process.

addons/func_godot/src/core/entity_assembler.gd
redxdev: That sort of undermines the intent, as if there were a new step after entity assembly this new signal would need to run after that new step (to closer mimic a deferred call). Intent is this runs effectively at the existing build_complete signal, though I have it a bit earlier to let timing metrics capture how long it takes to call the entity build complete methods.

redxdev:
@RhapsodyInGeek Hmm... not really sure how to square that. Like my original post said, deferred calls are problematic if func_godot is used during a headless import as the engine doesn't seem to guarantee that pending deferred calls run to completion when exiting... not sure if that's an engine bug, an oversight, or expected behavior so it's easier to fix here. Will have to think about this more.

redxdev:
I guess one option would be a separate non-deferred callback in addition to (instead of replacing) `_func_godot_build_complete`, though I don't love that solution either.

RisingThumb:
@RhapsodyInGeek are you 100% certain that this change breaks your project?

redxdev:
Another option could be to have multiple modes for this (as clunky as that would be). When running with a headless import the user can emit the "complete" signal manually after the build but before the import completes, while default behavior would be to emit the complete signal via a call_deferred. Just spitballing other options.

I'm also actively looking through godot's code to see if there's any way to force deferred calls to flush during an import but it doesn't seem like it.

It'd be good to verify this change is actually problematic though - I've got a whole list of hacky things I can try but I keep coming back to the fact that doing additional build operations after the build has technically completed just seems... wrong.

RhapsodyInGeek:
I'm not 100% sure, as I haven't had a chance to test this change. But I do know that the reason why I use `_build_complete()` is because I need the call to be deferred in order to complete.

An example being my `func_move` object. In the tick where the node is created, it calls `_func_godot_apply_properties()` and I set all of the variables and apply changes to the generated meshes. I then run `_func_godot_build_complete()` in order to generate and add multiple child nodes to the func_move, which can't have its children modified during `_func_godot_apply_properties()` (Godot complains and aborts the operations if I try).

I also run all my reparenting functions in `_func_godot_build_complete()`.

redxdev:
Was away for a bit so only just came back to this - I still feel like something is pretty wrong with the design of building a map if the build is generally not considered to be async but also isn't really complete when `build()` returns. Needing to defer an operation to later feels like the realm of user-code, not something a tool that for all intents and purposes finishes synchronously should be providing callbacks for. Though I could see an argument made for having a _new_ callback that uses my method in this PR so that backwards compatibility is retained, though I'll have to think on a new name for it...

I'm also bit confused as to what could be getting locked on a node @RhapsodyInGeek - would be interesting to know exactly what the error is, because it'd be very strange to not be able to add additional nodes to the map during a build when that's what the map build itself is doing.

RhapsodyInGeek:
> would be interesting to know exactly what the error is, because it'd be very strange to not be able to add additional nodes to the map during a build when that's what the map build itself is doing.

Can't remember the exact error. I just know I couldn't add child nodes the same frame the would-be parent nodes were added to the SceneTree. `_func_godot_build_complete()` was added to the build process as a deferred call specifically for that reason: to ensure that node modifications were applied properly before performing a second modification.

I remember we did first try to just call `_func_godot_build_complete()` the same frame in a separate loop, but the aforementioned SceneTree restrictions pushed us to make it a deferred call instead.

Just because you feel the design is wrong doesn't mean that it is; a deferred call after the map building has completed felt correct enough to the team that we put it in the plugin on day one. Just simply differing design philosophies, and we went with the one that works for the work flow that does the job the way we need it to.

It sounds like this PR is stemming from attempts to import maps as packed scenes?

redxdev:
> It sounds like this PR is stemming from attempts to import maps as packed scenes?

Correct, and unfortunately to my knowledge godot doesn't really give you any kind of control over when an import operation is considered complete, therefore making the use of deferred calls incredibly unsafe (even outside of my command-line issues, I wonder if a deferred modification to the packed scene would even get saved as part of the import).

> Just simply differing design philosophies, and we went with the one that works for the work flow that does the job the way we need it to.

I have two issues with this which is why I brought it up. First, I feel like it breaks the implied contract of the API - nothing about `FuncGodotMap.build()` implies that completion is asynchronous. The fact that the completion callback is deferred is not an obvious aspect of how func_godot works.

Secondly, making a separate deferred call for when the build is complete is already trivial without func_godot providing that itself (since you can do `_my_custom_complete_func.call_deferred()` - one extra line over func_godot providing the callback for you). It is a bit less trivial to get a non-deferred callback with func_godot's current implementation as you would need to walk up the node tree to find your parent `FuncGodotMap` to connect the completion signal.

If nothing else, I think having a non-deferred callback is important. And maybe I simply haven't run into enough of a reason to have a deferred callback provided out of the box - but even then I'd argue that the current naming is misleading.

RhapsodyInGeek:
> I wonder if a deferred modification to the packed scene would even get saved as part of the import

I have no idea, but the question arose in my mind as well. The main thing to remember though is that FuncGodot was never meant to import maps as scene files: this is something that came from several community members after its release but was never on our roadmap and is simply not how it was designed to be used. While we're all for adding features after the fact (including maps-as-packed-scenes support should a sufficiently satisfying implementation get a PR), sometimes there are going to be limitations as a result of attempting to use software in unaccounted for ways.

> First, I feel like it breaks the implied contract of the API - nothing about FuncGodotMap.build() implies that completion is asynchronous

It used to be asynchronous actually, as we originally used the build code from Qodot (cleaned up best we could but still Qodot) which completed parts of the build process as tasks that would take several frames to complete. We only changed the build process in the Core Rewrite to run almost in one frame, with some consideration towards splitting it up across more. Ultimately it was just simpler for us to do it the way it currently is. But frankly, I never took `FuncGodotMap.build()` to imply that it was async or not. I only took it to imply that it was building the map file, and nothing more than that. I'm not sure what else it would actually mean.

> The fact that the completion callback is deferred is not an obvious aspect of how func_godot works.

You could make the same argument for `_func_godot_apply_properties(properties)`, honestly, in the sense of whether or not it's obvious that you're supposed to add that function to your scripts and that you need to make your script a tool script. Both methods are documented in the manual, they're both plainly visible in the code. Neither have instructions outside of those two locations. Feels like a non-argument to me.

Truth is there are so many aspects to the plugin that are not obvious, which is why I went so hard on the documentation and the team went so hard on trying to make sure all the code and properties are heavily commented.

We named it `_func_godot_build_complete()` because it runs upon the map building being complete. That's how we view it, and it's fine if you disagree but I'll admit you're the first one to have raised any contention over the naming. I think it's just differing views on what accounts for obvious and accurate.

redxdev:
> The main thing to remember though is that FuncGodot was never meant to import maps as scene files

Yeah I get this, and I'm really not trying to impose it as a core design goal - but func_godot's design is such that it can and does work with only minor changes and if setup properly results in quite a nice workflow (though I'm still finding the occasional rough edge myself... mostly due to godot rather than func_godot).

> But frankly, I never took FuncGodotMap.build() to imply that it was async or not. I only took it to imply that it was building the map file, and nothing more than that. I'm not sure what else it would actually mean.

The problem is that there's nothing that tells you this, in the API or the docs, until you test it yourself or look into how the completion callbacks are implemented. Even ignoring my use-case, I can easily imagine someone calling build and then expecting to be able to interact with the finished result only to find that they now have their post-build code executing before any of the callbacks. Actually, it's a bit worse than that - `build_complete` will be emit a full frame before any completion callbacks happen which is unintuitive to say the least.

> I'll admit you're the first one to have raised any contention over the naming. I think it's just differing views on what accounts for obvious and accurate.

It's definitely nitpicky on my part - but I'm not looking _just_ to change the name of the completion callback, because my main point is that at the very least that there should be a separate non-deferred completion callback without having to do something less-than-ideal like `(get_parent() as FuncGodotMap).build_complete.connect(my_non_deferred_completion_func, CONNECT_ONE_SHOT)`. Which I'm happy to change this PR into a solution for, just want to see if that's something you'd even consider first. My point with the naming is more that if a new callback is added, "complete" becomes a bit confusing between the two.

RisingThumb:
> Correct, and unfortunately to my knowledge godot doesn't really give you any kind of control over when an import operation is considered complete, therefore making the use of deferred calls incredibly unsafe (even outside of my command-line issues, I wonder if a deferred modification to the packed scene would even get saved as part of the import).

FYI, this change is very specific to running in headless mode from the console as `--import --headless` as you specified originally. I just tested it out in 4.7.1 and _func_godot_build_complete() runs when importing as a scene normally in the editor, so there is some difference between running it headlessly and running it in-editor. This _**may**_ be a bug with Godot, so it is worth looking upstream for more information about it wrt to --headless before pursuing this change, and if not, file a bug report about this difference with some sort of minimum project.

From my perspective, the only approach that doesn't introduce breaking changes is a flag that lets you toggle between the behaviours defaulting to the current behaviour. Being pragmatic, I think it's either this compromise, or you keep this change as a personal patch on your personal fork of func_godot. The compromise could probably be revisited if func_godot has a v2 that makes a lot of breaking changes. In an ideal world though, we find out if this is a bug with Godot, a bug report gets made upstream and it gets fixed.

Also a minor thing... there are edge cases with scene importing even in the editor. The only edge case I've encountered, is that there's no scene tree during the import process for importing as a scene. Either that, or I'm stupid(very possible) and wasn't able to get the scene tree while importing as a scene- but this means you have to engineer an alternative for when that's relevant. Mostly relevant for anything involving groups where you'd normally do `get_tree().get_nodes_in_group()` and equivalent.

redxdev:
> This may be a bug with Godot, so it is worth looking upstream for more information about it wrt to --headless before pursuing this change, and if not, file a bug report about this difference with some sort of minimum project.

I agree, I just haven't had the time to do it and fixing it here (and in my own fork) was a bit more prudent to getting things working. I'll probably file something upstream at some point, though I have a feeling this might be inherent behavior to the engine shutting down when deferred calls are pending. I still worry that a deferred call might not be guaranteed to run by the time the editor (in headless import mode or not) tries to save out the imported asset, though.

> From my perspective, the only approach that doesn't introduce breaking changes is a flag that lets you toggle between the behaviours defaulting to the current behaviour.

Or adding a second callback - we can have both available for projects that might need both. It just runs into the naming confusion that I've mentioned, which isn't the end of the world or anything.

> Also a minor thing... there are edge cases with scene importing even in the editor. The only edge case I've encountered, is that there's no scene tree during the import process for importing as a scene.

Yeah like I said I'm still running into the occasional rough edge. The solution to this one is actually to create a temporary scene tree that isn't fully initialized, add a `FuncGodotMap` node into it, build, then remove it from the temp tree to save it out. It kinda just works as long as nothing tries to access functionality that would require a fully initialized scene tree.

...which I did need, as I wanted to build navmesh on import. But that's a me problem (and one that I solved by temporarily adding the nodes to the editor's scene tree, which works fine but is hacky and I fully expect it to break on some future godot update haha).

redxdev:
Actually, I'm now wondering if it's possible to `await` within an import plugin... because I could pretty easily create a signal for when all of the completion callbacks are done, await it, and then it's guaranteed that everything is done. Assuming it'd work, that'd retain the current deferred call behavior while making my use-case possible.

Should only require some minor modifications to func_godot, going to test it real quick...

redxdev:
To my surprise, it works? Was not expecting `await` to work properly within a supposedly synchronous call to `EditorSceneFormatImporter._import_scene`. I guess the engine side allows deferred calls to execute while waiting on a coroutine.

I'm going to update the PR though I think I can now handle this entirely from user-code with no modifications to `func_godot`, so this PR isn't really necessary anymore.

redxdev:
...or not. Found some wild behavior when deleting imported data where the editor somehow got into an infinite loop of trying to import the same assets over and over again. I take it that the engine isn't really built to await in weird spots like this.

RhapsodyInGeek:
> > But frankly, I never took FuncGodotMap.build() to imply that it was async or not. I only took it to imply that it was building the map file, and nothing more than that. I'm not sure what else it would actually mean.
>
> The problem is that there's nothing that tells you this, in the API or the docs, until you test it yourself or look into how the completion callbacks are implemented.

Your reply doesn't seem to be in response to what you quoted? If it is, this is the FuncGodotMap reference's `build()` description from the online docs:

> Cleans the map node of any children, then initiates and oversees the build process.

This is what it is in the plugin script's autodocs + comments read for the `build()` method:

> Builds the global_map_file. If not set, builds the local_map_file. First cleans the map node of any children, then creates a FuncGodotParser, FuncGodotGeometryGenerator  and FuncGodotEntityAssembler to parse and generate the map.

Nothing in here implies anything that you keep saying about synchronicity. I still fail to see any compelling argument as to how it breaks any contracts with the API. It does what it says.

However, the phrasing in your response sounds more like continued disagreement with how the `_func_godot_complete()` callback is not obvious about it being a deferred call. You keep making the claim that there's nothing that tells you how it behaves.

However, in the documentation, under *Forge Game Data > Modifying Entities On Map Build*, it literally states that it is a deferred call. It even shares the code snippet from Entity Assembler above the following description:

> If you decide you want to make modifications to your generated entities on map build using any of the following methods, you'll need to make sure that your entity's script is set as a `@tool` script.
> 
> The first method call is to `_func_godot_apply_properties(entity_properties: Dictionary)`. Because `_func_godot_apply_properties` passes the entity properties in the function, you do not need to declare a `func_godot_properties` dictionary in your node script; instead, just apply your properties as needed in this callback.
> 
> After every entity has called `_func_godot_apply_properties`, FuncGodot will cycle through all of them again and attempt to call `_func_godot_build_complete()` as a deferred call. Since we're calling this method deferred after every entity has already called the previous method, we can guarantee any nodes generated by our entities in `_func_godot_apply_properties` can be referenced during `_func_godot_build_complete`.

In both the manual and the Godot autodocs + code comments for `FuncGodotEntityAssembler.apply_entity_properties()` also state plainly that this is the behavior:

> Converts the String values of the entity data's properties Dictionary to various Variant formats  based upon the [FuncGodotFGDEntity]'s class properties, then attempts to send those properties to a func_godot_properties Dictionary  and an _func_godot_apply_properties(properties: Dictionary) method on the node. A deferred call to _func_godot_build_complete() is also made.

The argument that it's undocumented behavior is, quite frankly, absurd. I understand that you disagree with the implementation because it's causing you problems in this particular use case, but you don't need to keep stating things that are untrue to make your case.

redxdev:
> Nothing in here implies anything that you keep saying about synchronicity. I still fail to see any compelling argument as to how it breaks any contracts with the API. It does what it says.

It's not that it states anything explicit about (a)synchronicity, but I was operating under the assumption that function calls tend to be synchronous by default unless either stated otherwise or when their use must obviously be asynchronous, and this breaks that. Of course that's my own assumption, and if you disagree with that then that's the end of it.

> but you don't need to keep stating things that are untrue to make your case.

I admit that I missed that one bit of documentation. I'm not trying to argue just to be annoying or to be right, but because my experience is that most libraries are explicit about asynchronicity because in the vast majority of cases a function call is synchronous, and being asynchronous ends up documented explicitly because it changes how you have to handle a call. It's not just about my use-case, it's about the general API design and documentation.

I don't want to keep hammering on this point though, we can agree to disagree on how such an API should be designed - it doesn't really change my needs and I'm happy to either continue maintaining my own fork with my changes or to contribute back the additional callback I mentioned.

RisingThumb:
> I don't want to keep hammering on this point though, we can agree to disagree on how such an API should be designed - it doesn't really change my needs and I'm happy to either continue maintaining my own fork with my changes or to contribute back the additional callback I mentioned.

Personally, I think you should look upstream to see if it's an issue with Godot, before it's something to consider for func_godot. Any of the reasonable ways to add this to func_godot add jank for what I think should just work the same as it does in the editor.

redxdev:
I'm not sure how a second build complete callback adds jank? I'm not saying it would replace the existing one in any circumstance (not anymore, at least - ignore what this PR currently does), it would be in addition to rather than a replacement for. I get that it's not the cleanest solution either, but even if things were fixed upstream I think there'd still be other jank caused by relying on deferred calls that execute after an import plugin has supposedly finished its work.

PR #266, opened 2026-08-10
Allow _func_godot_build_complete to optionally take the entity_properties dictionary as an argument

redxdev:
This lets `_func_godot_build_complete` be defined on entities in either of these ways:

```gdscript
func _func_godot_build_complete() -> void:
  # do some stuff
  pass

func _func_godot_build_complete(entity_properties: Dictionary) -> void:
  # do some stuff, but with the entity's property dictionary
  pass
```

I feel that the latter is more generally useful, but for backwards-compatibility this should work with the former as well. My reasoning for this change is that I've found that I often need to lookup information about other nodes after the map build is complete (since they might not have been created by the time `_func_godo_apply_properties` is called), but the information I need to know what to do is contained within the entity properties dictionary which isn't passed to `_func_godot_build_complete`.

One workaround is to define a `func_godot_properties` property for the node, but that's a bit of a heavy-handed change as most of the time I don't want the full property dictionary stored as part of the node - I just need a couple of properties and I don't need them to persist.

### DIFF
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -283,9 +283,14 @@ func apply_entity_properties(node: Node, data: _EntityData) -> void:
 	
 	if node.has_method("_func_godot_build_complete"):
 		var complete_method : Callable = node.get("_func_godot_build_complete")
-		if complete_method.get_argument_count() > 0:
-			# Allow _func_godot_build_complete to optionally have properties passed to it
-			complete_method = complete_method.bind(properties)
+		match complete_method.get_argument_count():
+			0:
+				pass
+			1:
+				# Allow _func_godot_build_complete to optionally have properties passed to it
+				complete_method = complete_method.bind(properties)
+			_:
+				push_error("Entity %s script has a _func_godot_build_complete function with an incorrect number of arguments (expected 0 or 1) - the call may fail" % node.name)
 		complete_method.call_deferred()
 
 ## Generate a [Node] from [FuncGodotData.EntityData]. The returned node value can be [code]null[/code], 
--- a/addons/func_godot/src/core/entity_assembler.gd
+++ b/addons/func_godot/src/core/entity_assembler.gd
@@ -282,7 +282,11 @@ func apply_entity_properties(node: Node, data: _EntityData) -> void:
 		node.call("_func_godot_apply_properties", properties)
 	
 	if node.has_method("_func_godot_build_complete"):
-		node.call_deferred("_func_godot_build_complete")
+		var complete_method : Callable = node.get("_func_godot_build_complete")
+		if complete_method.get_argument_count() > 0:
+			# Allow _func_godot_build_complete to optionally have properties passed to it
+			complete_method = complete_method.bind(properties)
+		complete_method.call_deferred()
 
 ## Generate a [Node] from [FuncGodotData.EntityData]. The returned node value can be [code]null[/code], 
 ## in the case of [FuncGodotFGDSolidClass] entities with no [FuncGodotData.BrushData] entries.


redxdev:
To illustrate what I'm trying to solve, in case it wasn't clear, I have this part of a script on a node created by func_godot:

```gdscript
func _func_godot_apply_properties(entity_properties: Dictionary) -> void:
	if entity_properties.has("idle_path_target"):
		var target_name := entity_properties["idle_path_target"] as String
		if not target_name.is_empty():
			_apply_idle_path.call_deferred(target_name)

func _apply_idle_path(target_name: StringName) -> void:
	idle_path = InfoWaypoint.find_targets(get_parent(), target_name)
```

which works _fine_, but it's not ideal that I need to deal with some deferred call handling myself when the build_complete callback would work fine if not for the missing entity properties dictionary. With this PR, I can change it to this single function:

```gdscript
func _func_godot_build_complete(entity_properties: Dictionary) -> void:
	if entity_properties.has("idle_path_target"):
		var target_name := entity_properties["idle_path_target"] as String
		if not target_name.is_empty():
			idle_path = InfoWaypoint.find_targets(get_parent(), target_name)
```

sinewavey:
In the past, we actually _did_ use the func_godot_properties dictionary member. I do wish Godot exposed a way to check the type of the parameter of callables, though.

However, I think #264 might supersede this, as effectively it would take advantage of an already built-in dictionary for each node - would like to hear @RhapsodyInGeek 's thoughts here

redxdev:
I saw that PR and thought about using it instead for what I'm working on, but it still has the downside of requiring the additional storage of entity properties when you don't actually need them at runtime/after the map build is complete. I think both have their use-cases, and while entity properties inside the node meta _can_ be used to solve this same issue I think this solution is cleaner (but node meta still has its place and should be considered separately).

RisingThumb:
@redxdev consider the following:

```
@export car func_godot_properties: Dictionary

func _func_godot_build_complete() -> void:
    func_godot_properties.get("targetname", "")
    ...
    func_godot_properties = {}
```
While this should work currently, it's clunky, so I'm in support of this change you've suggested.

If a user defines the following:
```
func _func_godot_build_complete()-> void:
    print("1")
func _func_godoy_build_complete(ents: Dictionary)->void:
    print("2")
```
The 2nd function, the one accepting arguments is always called right?

@sinewavey #264 doesn't solve this as instead of func_godot_properties being a variable it becomes a meta property and gets stored all the same. If anything it becomes slightly harder to deal with as you're now doing meta properties functions to get and set it which would make things clunkier than what I showed above

redxdev:
> While this should work currently, it's clunky, so I'm in support of this change you've suggested.

Yeah, that bit of clunkiness and the extra property is what I'm trying to avoid with this change. It's just not necessary when the property dictionary can be passed in just fine.

> The 2nd function, the one accepting arguments is always called right?

Correct. Though... gdscript doesn't have method overloads right? So you wouldn't be able to have that situation anyway. It does mean that if a parent class defined `_func_godot_build_complete()` then you wouldn't be able to make a subclass that defined `_func_godot_build_complete(entity_properties: Dictionary)` but that seems like a minor and easily-fixable issue in my view.

sinewavey:
> https://github.com/func-godot/func_godot_plugin/pull/264 doesn't solve this as instead of func_godot_properties being a variable it becomes a meta property and gets stored all the same. If anything it becomes slightly harder to deal with as you're now doing meta properties functions to get and set it which would make things clunkier than what I showed above

Yeah, that's fair. I figured you can always clear metadata after use, and you wouldn't have to define an empty class member (an object still has the empty metadata dict, after all) - but it's still a bit clunky and might require a bit more specific cleanup than an optional arg in a function. While I do think the GDscript support for this kind of thing is a bit shaky, overall I'm in favor of this change and will give it a test.

RisingThumb:
> Correct. Though... gdscript doesn't have method overloads right? So you wouldn't be able to have that situation anyway. It does mean that if a parent class defined 

I can't remember off the top of my head 😅

RhapsodyInGeek:
So when we were doing the fork from Qodot to FuncGodot, I had actually originally tried to pass `func_godot_properties` to `_func_godot_build_complete()`. The problem I had kept running into was that I couldn't guarantee that the properties Dictionary would still exist when `_func_godot_build_complete()` was called. This was through both a one shot callback IIRC and the current deferred call method. The problem was that the properties dictionary was effectively local to the entity construction functions and Godot considered the references to it to be non-existent as soon as the entity assembler function was finished.

I haven't attempted to do it since we've rewritten the entire Core scripts, which no longer work at all like the old Qodot code, so if the properties Dictionary is persisting long enough to be passed along to the _build_complete() call then I'm all for this. The implementation looks clean and maintains backwards compatibility without users having to do anything, so it all looks good to me.

redxdev:
I'd have to dig into Godot's source a bit, but I'd expect the dictionary to stick around because `bind()` should be retaining a reference to it. I guess some kind of weird interaction with dictionary because it's not actually RefCounted? Not sure. I've been using this change locally for a bit and haven't run into any issues, though.

RhapsodyInGeek:
Dictionaries and Arrays are weird because they're passed by reference despite not being RefCounted, which is probably why they disappear as soon as the object that originally created them disappears (the object in this case being the callable the properties dictionary was created in). 

In the old Qodot system func_godot originally used, it had a whole bunch of methods that it ran through on some big loop, really overdesigned machinery, and IIRC the properties data was gathered and parsed during the entity assembly function.

In the core rewrite @sinewavey and I did, all of that data is gathered and interpreted at the start during the Parser stage and applied directly to the EntityData objects, so those Dictionaries persist as long as the RefCounted EntityData does AFAIK.

The safest thing would probably be to try to pass a `duplicate(true)` of the properties, but I don't know that it's necessary if the parse data is persisting on the map node?

redxdev:
fwiw I'm not persisting the dictionary anywhere - I don't have a func_godot_properties variable on nodes using this and it works fine. Would probably be good to have more testing on it, but if nothing else this shouldn't break existing projects because no one can even have a function that takes the dictionary yet.

I can add a duplicate to it but I'm also not sure that'd do anything but waste a bit of memory.

sinewavey:
As far as functionality, I tested this morning for the typical workflow (importing via build on the map node in editor)

The only thing I can think of - excluding the value type semantics of the properties dictionary) is the result of getting the function's argc being greater than 1, i.e. someone defines `_build_complete(arg0: Dictionary, arg1: Variant, ...). 

That's definitely a case of end user error, and we might want to have something check/block the call in that case, though Godot will still give you an error itself from the failed call.

I think if we can figure out a solid answer on the dictionary lifetime the rest looks good to me.

redxdev:
> That's definitely a case of end user error, and we might want to have something check/block the call in that case, though Godot will still give you an error itself from the failed call.

Surfacing it with an obvious error that the signature is incorrect is probably a bit nicer - I'll add that.

> I think if we can figure out a solid answer on the dictionary lifetime the rest looks good to me.

I'll do a bit of digging today to see if I can't figure out the exact rules around dictionary lifetime.

redxdev:
As far as I can tell, dictionaries (and arrays, and some other types) _are_ refcounted - they just don't implement it via `RefCounted` for whatever reason. See https://github.com/godotengine/godot/blob/master/core/variant/dictionary.cpp#L287

So what I'm doing should be safe, though I'm still going through the code a bit to verify.

redxdev:
Took a bit more time to look through godot's source and yeah I don't see why this shouldn't work consistently (and it has been working fine for me on my own project).

sinewavey:
Yeah, it does seem that should be OK.

Thanks!

sinewavey:
Ultimately we do think the feature will make it in, but after some talk with @RhapsodyInGeek there might be some interplay with the property def resources, so holding off for now

redxdev:
My property resource PRs work fine with this as well (I'm using a custom branch that includes all of the work I've submitted over the past few weeks - the PRs are adapted from that) but if you have further plans I get it. Happy to help if anything else is needed.

PR #267, opened 2026-08-11
Add a 'FuncGodotChoicesProperty' resource as an alternate (and simpler) way to define choices properties

redxdev:
This is based on the same technique I used in #261 for defining special property types.

This adds a new `FuncGodotChoicesProperty` resource that can be used instead of the somewhat clunky method of defining choices properties via a dictionary. Now you define the property default as the new `FuncGodotChoicesProperty` type (select `Object` as the type, then make a new `FuncGodotChoicesProperty` or select an existing resource) which is _much_ simpler to setup and doesn't require using the property descriptions to set a default value.

image

image

An extra benefit of this method is that you can create shared choices resources if you have multiple properties that have the same choices (like for shared enums).

### DIFF
--- a/addons/func_godot/src/core/parser.gd
+++ b/addons/func_godot/src/core/parser.gd
@@ -189,8 +189,12 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 			var prop_string = entity.properties[property]
 			if property in def.class_properties:
 				var prop_default: Variant = def.class_properties[property]
+
+				var expected_type := typeof(prop_default)
+				if prop_default is FuncGodotChoicesProperty:
+					expected_type = typeof((prop_default as FuncGodotChoicesProperty).default_value)
 				
-				match typeof(prop_default):
+				match expected_type:
 					TYPE_INT:
 						properties[property] = prop_string.to_int()
 					TYPE_FLOAT:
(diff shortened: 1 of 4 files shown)


addons/func_godot/src/fgd/func_godot_choices_property.gd
```diff
@@ -0,0 +1,9 @@
+class_name FuncGodotChoicesProperty
+extends Resource
+## Special resource used to indicate that an FGD entity property is of type "choices".
+
+## The default value for this property.
+@export var default_value: Variant
+
+## The available choices for this property.
+@export var choices: Dictionary[Variant, String]
```
redxdev, line 9: I implemented the "values" as variants here because I wasn't sure what various editors actually support for choices... originally I used ints, but I guess strings might also work in some (all? none?)

addons/func_godot/src/core/parser.gd
```diff
@@ -189,8 +189,12 @@ func parse_map_data(map_file: String, map_settings: FuncGodotMapSettings) -> _Pa
 			var prop_string = entity.properties[property]
 			if property in def.class_properties:
 				var prop_default: Variant = def.class_properties[property]
+
+				var expected_type := typeof(prop_default)
+				if prop_default is FuncGodotChoicesProperty:
+					expected_type = typeof((prop_default as FuncGodotChoicesProperty).default_value)
```
redxdev, line 195: Noting that the choice of property type is based on the default value variant. I don't know if choices support mixed-types (if they support various options for types beyond integers at all) but if they do I'll have to figure out another solution here.

addons/func_godot/src/fgd/func_godot_choices_property.gd
RisingThumb: Strings and Ints are valid for this, and there's no type unions in Godot, so variant is the best you get

addons/func_godot/src/core/parser.gd
RhapsodyInGeek: I believe they do, as MAP key value pairs all end up as strings anyway and don't actually have type safety. It's actually the game itself that determines if an entity property is an incorrect type (by crashing on you) upon trying to assign it to a spawned entity's field variables in the original id Tech and GldSrc engines.

That said, just about every non-Array Variant type in Godot can be converted to any other non-Array Variant type so it should be fine and it's the direction I would take designing this, too.

PR #268, opened 2026-08-11
Better method of defining flags via a new FuncGodotFlagsPropertyEntry Resource

redxdev:
See also #267 and #261 for my applications of a similar technique for choices and target source properties.

For flags I opted to not create a top-level resource for the flag type (for now) but it'd be easy enough to add later. The main thing I wanted to do with this change is make it clear what the available fields are, instead of having to remember whether the name/value/default goes in the first/second/third index of an array. To that end, the "top-level" type of flags is still `Array`, but now you can make `FuncGodotFlagsPropertyEntry` resources instead of sub-arrays.

image


RhapsodyInGeek:
I was actually trying to develop a similar feature some time ago, but didn't come up with a solution that I really liked. I was originally looking for a catch-all `FuncGodotFGDPropertyType` resource, but obviously that would get a bit complicated pretty easily (not that the plugin isn't already complicated).

What I'd really like to see from something like this is instead of putting in the bit flag value (_4, 16, 32, 524288, etc..._) instead put the flag index (_1, 2, 3, 4... 24_ (max value that the map editors can see)).

It otherwise looks good to me, but I don't know if we're ready to merge this just yet, as once something like this is in it's not leaving, but it is the right direction that we've been interested in.

redxdev:
> What I'd really like to see from something like this is instead of putting in the bit flag value (4, 16, 32, 524288, etc...) instead put the flag index (1, 2, 3, 4... 24 (max value that the map editors can see)).

Ah yeah, I can change it to work like that. Actually, I can probably let it get set either way. Will add that.

> It otherwise looks good to me, but I don't know if we're ready to merge this just yet, as once something like this is in it's not leaving, but it is the right direction that we've been interested in.

Yeah, totally understandable. I have my own working copy of func_godot for my game so I'm just throwing these up as I make improvements for my own workflow :)

redxdev:
Ended up just exposing the "index" of the bitflag. Used 0-based indexing (`1 << index`, as opposed to `1 << (index - 1)`), though I was going back and forth on that because godot annoyingly does 1-based indexing for naming layers (i.e. the first render layer is layer 1, not layer 0)...

RisingThumb:
Only suggestion would be to add a warning if the editor is NetRadiantCustom and the key name is not "spawnflags" or something like that as NRC breaks if you do that. CC @sinewavey to confirm as our resident NRC user

redxdev:
Probably best for another PR? Doing so isn't really dependent on this change.

RhapsodyInGeek:
> Only suggestion would be to add a warning if the editor is NetRadiantCustom and the key name is not "spawnflags" or something like that as NRC breaks if you do that. CC @sinewavey to confirm as our resident NRC user

NRC doesn't need you to call them `spawnflags`, it will automatically call them that (same as JACK). However, you cannot have conflicting flag assignments like in other editors, otherwise NRC crashes. That's something I think would be unreasonable for us to try to catch all the different edge cases that could arise and we should simply educate users on instead.

addons/func_godot/src/fgd/func_godot_flags_property_entry.gd
```diff
@@ -0,0 +1,17 @@
...
+class_name FuncGodotFlagsPropertyEntry
+extends Resource
+## Special resource used to make setting flags as a property type on an entity easier.
+
+## The name of the flag
+@export var name: String
+@export var value := 0
+
+## The "index" of the bitflag. For example: index 0 is equal to a raw value of 1, index 2 is equal to a raw value of 2, index 3 is equal to a raw value of 4, index 4 is equal to a raw value of 8, etc.
+@export_range(0, 31, 1) var index := 0
```
RhapsodyInGeek: Export range should probably be 0 - 23, because as far as I know no Quake map editors support more than 24 bits. TrenchBroom and JACK only support 24 bits, I cannot recall if NRC is limited to that as well. Funny enough, the reason for this is that QuakeC doesn't actually have integers, only 32 bit floats, so only 24 bits are available for bitwise ops. This behavior was inherited by GldSrc games, as well as any other games inheriting from the Quake engine.

@sinewavey do you happen to know if NRC can support more than 24 spawnflags on a single entity?

redxdev: Yeah I wasn't sure if there were editors that supported more than 24 out there which is why I left it at 32 values, but I can change it to 24 values if there aren't any editors supporting that atm.

PR #269, opened 2026-08-19
Added engine path in TrenchBroomGameConfig

StepingForward:
This allows the user to set the engine for trenchbroom inside the TrenchBroomGameConfig file. Before you had to do it in Trenchbroom by going ```View->Preferences->YourGame->GamePath```. 

You can leave this field empty, if you still prefer doing it that way. 

This works by getting the ```Preferences.json``` file relative to the game folder and adding "Games/game_name/Path": "engine_path" attribute to it.

image


RhapsodyInGeek:
Unfortunately we can't use literal global paths on version controlled resources, only relative project paths. Otherwise it will break team projects, where various team members may not have the same tool installation or work folder paths on their own machines.

This is actually the reason the [FuncGodotLocalConfig](https://func-godot.github.io/func_godot_docs/FuncGodot%20Manual/pages/ref_local_config.html#LocalConfig) resource exists: for local literal paths that are not intended to be shared across multiple machines.

StepingForward:
Yeah, that makes sense, I didn't even think of it from the team perspective. Thanks for the quick answer.

PR #272, opened 2026-08-27
Add a way for the wound vertices to be overriden in an entity definition

RisingThumb:
This provides an optional script on an entity definition, providing a hook to override the placement of brush vertices during a func_godot build process.

This should be done during the func_godot build process and not on the users side in _func_godot_apply_properties or the build complete, as the new position of these vertices is relevant to setting up any new concave collisions, the UV2 generation(I think?) and the normals(especially relevant when phong shading is set up).

An example of a script that could be used with this allowing a flat brush that has been manually subdivided to have some waviness is as follows:
```
extends RefCounted

var mult := 5.0

func _func_godot_modify_vertex_wind(wind: PackedVector3Array, _entity: FuncGodotData.EntityData) -> void:
	var noisy: FastNoiseLite = load("res://func_godot/classes/func_/func_cave_wall/func_cave_noise.tres")

	for i in wind.size():
		wind[i].z += noisy.get_noise_3d(wind[i].x, wind[i].y, wind[i].z)*mult
```
Some users were already hacking func_godot in the wild to allow for this sort of change(see 0rangeyouglad in showcase-mapping as an example). The entity is also provided so users can get key-value pairs as they are probably relevant if you want to override it using some user-defined params.

I think this is kind of niche change but allows for natural organic terrain to be somewhat more easily done rather than manually changing each vertex in your map editor for when you would rather give up most of your control with it. I think this could also allow users to do automatic subdivisions(I haven't tested that idea out though and would require understanding how the verts are wound up which I don't know how they are done)?

The implementation for this is similar to how I implemented the property attacher stuff. I'm not sure about the names for variables in this change(happy to hear alternatives).

### DIFF
--- a/addons/func_godot/src/core/geometry_generator.gd
+++ b/addons/func_godot/src/core/geometry_generator.gd
@@ -106,7 +106,7 @@ func generate_base_winding(plane: Plane) -> PackedVector3Array:
 	winding.append(centroid + (right *  h) + (forward * -h))
 	return winding
 
-func generate_face_vertices(brush: _BrushData, face_index: int, vertex_merge_distance: float = 0.0) -> PackedVector3Array:
+func generate_face_vertices(brush: _BrushData, face_index: int, entity: _EntityData, vertex_merge_distance: float = 0.0) -> PackedVector3Array:
 	var plane: Plane = brush.faces[face_index].plane
 	
 	# Generate initial square polygon to clip other planes against
@@ -133,6 +133,12 @@ func generate_face_vertices(brush: _BrushData, face_index: int, vertex_merge_dis
 				merged_winding.append(cur_vtx)
 			prev_vtx = cur_vtx
 		winding = merged_winding
+	
+	if entity.definition.brush_winding_overrider:
+		var ex: RefCounted = entity.definition.brush_winding_overrider.new()
+		if ex.has_method("_func_godot_modify_vertex_wind"):
+			ex.call("_func_godot_modify_vertex_wind", winding, entity)
+
 
 	return winding
 
@@ -143,7 +149,7 @@ func generate_brush_vertices(entity_index: int, brush_index: int) -> void:
 	
 	for face_index in brush.faces.size():
 		var face: _FaceData = brush.faces[face_index]
-		face.vertices = generate_face_vertices(brush, face_index, vertex_merge_distance)
+		face.vertices = generate_face_vertices(brush, face_index, entity, vertex_merge_distance)
 		
 		face.normals.resize(face.vertices.size())
 		face.normals.fill(face.plane.normal)
--- a/addons/func_godot/src/core/geometry_generator.gd
+++ b/addons/func_godot/src/core/geometry_generator.gd
@@ -110,7 +110,7 @@ func generate_face_vertices(brush: _BrushData, face_index: int, entity: _EntityD
 	var plane: Plane = brush.faces[face_index].plane
 	
 	# Generate initial square polygon to clip other planes against
-	var winding: PackedVector3Array = generate_base_winding(plane)
+	var vertices: PackedVector3Array = generate_base_winding(plane)
 
 	for other_face_index in brush.faces.size():
 		if other_face_index == face_index:
@@ -118,29 +118,29 @@ func generate_face_vertices(brush: _BrushData, face_index: int, entity: _EntityD
 		
 		# NOTE: This may need to be recentered to the origin, then moved back to the correct face position
 		# This problem may arise from floating point inaccuracy, given a large enough initial brush
-		winding = Geometry3D.clip_polygon(winding, brush.faces[other_face_index].plane)
-		if winding.is_empty():
+		vertices = Geometry3D.clip_polygon(vertices, brush.faces[other_face_index].plane)
+		if vertices.is_empty():
 			break
 	
 	# Perform rounding and merge adjacent vertices that are equivalent
 	if vertex_merge_distance > 0:
-		var merged_winding : PackedVector3Array = PackedVector3Array()
-		var prev_vtx : Vector3 = winding[0].snappedf(vertex_merge_distance)
-		merged_winding.append(prev_vtx)
-		for i in range(1, winding.size()):
-			var cur_vtx : Vector3 = winding[i].snappedf(vertex_merge_distance)
+		var merged_vertices : PackedVector3Array = PackedVector3Array()
+		var prev_vtx : Vector3 = vertices[0].snappedf(vertex_merge_distance)
+		merged_vertices.append(prev_vtx)
+		for i in range(1, vertices.size()):
+			var cur_vtx : Vector3 = vertices[i].snappedf(vertex_merge_distance)
 			if prev_vtx != cur_vtx:
-				merged_winding.append(cur_vtx)
+				merged_vertices.append(cur_vtx)
 			prev_vtx = cur_vtx
-		winding = merged_winding
+		vertices = merged_vertices
 	
-	if entity.definition.brush_winding_overrider:
-		var ex: RefCounted = entity.definition.brush_winding_overrider.new()
-		if ex.has_method("_func_godot_modify_vertex_wind"):
-			ex.call("_func_godot_modify_vertex_wind", winding, entity)
+	if entity.definition.modify_vertices_script:
+		var ex: RefCounted = entity.definition.modify_vertices_script.new()
+		if ex.has_method("_func_godot_modify_vertices"):
+			ex.call("_func_godot_modify_vertices", vertices, entity)
 
 
-	return winding
+	return vertices
 
 func generate_brush_vertices(entity_index: int, brush_index: int) -> void:
 	var entity: _EntityData = entity_data[entity_index]
--- a/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
+++ b/addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
@@ -61,6 +61,11 @@ var prefix: String = ""
 ## This can be used to programmatically generate class properties when you need it
 @export var entity_extension_script: Script
 
+## Optional script where `_func_godot_modify_vertex_wind` will be called during a build.
+## Provides the Vertices winding as a PackedVector3Array and the entity as a FuncGodotData.EntityData as arguments
+## These can be used to programmatically modify brush vertices during a build.
+@export var brush_winding_overrider : Script
+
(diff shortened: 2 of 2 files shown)


addons/func_godot/src/core/geometry_generator.gd
```diff
@@ -133,6 +133,12 @@ func generate_face_vertices(brush: _BrushData, face_index: int, vertex_merge_dis
 				merged_winding.append(cur_vtx)
 			prev_vtx = cur_vtx
 		winding = merged_winding
+	
+	if entity.definition.brush_winding_overrider:
```
Ralane: Would it make sense to make `_func_godot_modify_vertex_wind` return an array of callables (or perhaps some other approach like checking for a function naming scheme)?

That way we could move the existing vertex merge handling into one, as well as support multiple steps and reordering.

Might be overkill since there probably won't be many use cases for multiple vertex modifying steps within a pipeline, but still worth considering

RisingThumb: I think currently there's 2 points you can hook into. Before the vertex snapping and after the vertex snapping, so an array of callables would be overengineered when 2 functions would be sufficient.

I would like to hear a case for when pre-vertex snapping is worth hooking into though. The only thing I can think of where you'd want that over post-vertex snapping is when you want to snap all vertices to some position, but most map editors already provide that. When vertex snapping is disabled, they would be functionally equivalent hooks.

Ralane: Fair enough, yeah the more I think about it I can't imagine a use case for any sort of standard game mapping that'd need it

sinewavey:
Mirroring some thoughts here I mentioned in the Discord -- I think it's worth considering the naming here as this would allow overriding the generated vertices, not the brush winding process itself. 

Nitpicking aside, this is a use case I hadn't previously considered (and quite like!)

RhapsodyInGeek:
This feels strangely insufficient in its specificity to me. I like it, but I feel like there's something wrong in the implementation.

Echoing @sinewavey naming seems off for the same stated reasons.

RisingThumb:
@sinewavey mirroring your thoughts off of the Discord about vertex winding being openGL-wind order specific, does this mean the function `generate_face_vertices` should have the variable `winding` renamed to something else? I'm not sure if that is an OpenGL vertex wind? If it is, I'm not 100% convinced the hook function shouldn't include the word winding if winding is an OpenGL wind because if you know the winding order(I don't) you can probably do automatic subdivisions with it?

The name for the function does feel off to me still. I'll sleep on it.

> I like it, but I feel like there's something wrong in the implementation.

@RhapsodyInGeek I was considering the possibility that all the entity-specific hooks should be in a single script for entity hooks, however I'm unsure of other places where entity hooks would be needed. Class property attacher could also possibly be included as a function in that script too.

sinewavey:
To elaborate a little more: it probably shouldn't be called winding in the generate vertices function. It seems however that the clip polygon process does return the vertices already wound per openGL standards (so the tris face outward) but wasn't something I was sure if it was guaranteed - hence still performing it later.

It's probably more accurate for that array `winding` should be named `vertices` and the hook here to be something like `post_vertex_modifier`.

I'll have to look into it a bit later tomorrow to check when the vertices are generated and how overriding here might affect vertex normal/tangent/UV output, because that may also then be incorrect and require handling

RisingThumb:
> It's probably more accurate for that array winding should be named vertices and the hook here to be something like post_vertex_modifier.

I see, that being the case then yep, I'll rename the variable and then change the function name accordingly to be about the vertices. I am then wondering now if this is the incorrect place for automatic subdivision to be done?

> I'll have to look into it a bit later tomorrow to check when the vertices are generated and how overriding here might affect vertex normal/tangent/UV output, because that may also then be incorrect and require handling

On a cursory glance, I believe all of this happens in the generate_entity_surfaces function which happens right before unwrapping UV2s and well after the vertices are built looking at the build() function, but I'm not sure if they are using the vertices from the brush or the generated vertices for the mesh. On a brief inspection it looks like it uses face.vertices for calculating the normals, and we get those vertices earlier with `face.vertices = generate_face_vertices(...)`. I'll take a closer look tomorrow, but I don't think it is affected.

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
```diff
@@ -61,10 +61,10 @@ var prefix: String = ""
 ## This can be used to programmatically generate class properties when you need it
 @export var entity_extension_script: Script
 
-## Optional script where `_func_godot_modify_vertex_wind` will be called during a build.
-## Provides the Vertices winding as a PackedVector3Array and the entity as a FuncGodotData.EntityData as arguments
+## Optional script where `_func_godot_modify_vertices` will be called during a build.
+## Provides the Vertices as a PackedVector3Array and the entity as a FuncGodotData.EntityData as arguments
 ## These can be used to programmatically modify brush vertices during a build.
-@export var brush_winding_overrider : Script
+@export var modify_vertices_script : Script
```
RisingThumb, line 67: This could possibly be renamed to be a general purpose hooks script for this entity, but I'm unsure where else in the codebase it is useful to hook in for specifically that entity definition

addons/func_godot/src/core/geometry_generator.gd
```diff
@@ -110,37 +110,37 @@ func generate_face_vertices(brush: _BrushData, face_index: int, entity: _EntityD
...
+		var prev_vtx : Vector3 = vertices[0].snappedf(vertex_merge_distance)
+		merged_vertices.append(prev_vtx)
+		for i in range(1, vertices.size()):
+			var cur_vtx : Vector3 = vertices[i].snappedf(vertex_merge_distance)
 			if prev_vtx != cur_vtx:
-				merged_winding.append(cur_vtx)
+				merged_vertices.append(cur_vtx)
 			prev_vtx = cur_vtx
-		winding = merged_winding
+		vertices = merged_vertices
```
RisingThumb, line 135: All changes above this line are renaming merged_winding to merged_vertices and winding to vertices.

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
sinewavey: I could see the face normal/tangents also being something worth having a hook

RisingThumb: You raise a good point, I'll rename it to a general purpose entity-hooks script. Would you like me to add the hook for normal/tangents in this change? Or raise an issue after this for adding a hook that does that?

sinewavey: Yeah, it's probably best to include them now as potentially modifying the vertices could end up requiring the normals to be adjusted

RhapsodyInGeek:
> Class property attacher could also possibly be included as a function in that script too.

I think the class property attacher is best left where it is. The thing about the script property with that method is that it specifically modifies the FGD, not the build process. If someone wants to combine their methods into one script, then they can just reuse the same script across multiple properties.

RhapsodyInGeek:
That said, I think maybe we should rename this script property `Node Build Extension Script`. That way it can be reused across multiple build process hooks, especially since you're always checking the appropriate method exists anyway.

addons/func_godot/src/fgd/func_godot_fgd_entity_class.gd
RisingThumb: FYI, the normals get calculated based on what the vertices are after this, but yeah, I can see about adding a hook for the normals/tangents part since I can see that potentially being useful to some people.

Perhaps a hook for the normals also would allow alternatives to phong and... whatever the regular hard edge normals is called?

sinewavey: The normals right now are just unsmoothed and set to the plane normal they belonged to. Right after the face vertices are generated: 

```
face.normals.resize(face.vertices.size())
face.normals.fill(face.plane.normal)
```

addons/func_godot/src/core/geometry_generator.gd
RisingThumb: @sinewavey I did some playing around with this to see if automatic subdivision was possible. Just plain adding vertices to this list of vector 3s doesn't work, so the order of vertices does matter here so I'm now uncertain about calling it just vertices. I may have misunderstood this you posted earlier

> To elaborate a little more: it probably shouldn't be called winding in the generate vertices function. It seems however that the clip polygon process does return the vertices already wound per openGL standards (so the tris face outward) but wasn't something I was sure if it was guaranteed - hence still performing it later.
>
>It's probably more accurate for that array winding should be named vertices and the hook here to be something like post_vertex_modifier.

Would be interested in hearing your thoughts on this?

Ralane: After looking into it a bit for my own project, I think that although it may be technically possible to do automatic subdivision here at the vertex level, it would be much cleaner to do it by modifying the brush list rather than the vertex list.

That way, vertices for the newly generated brushes will be created automatically and handled the same as any other vertex (which may include for example vertex modification/snapping)

RisingThumb: Do share if you figure out a solution to it. WIth this issue, I was considering seeing if I can add an override on the faces, because I'm wondering if the faces could be split up this way. I haven't looked much into that approach though

Ralane: Oh that's a good idea, face modification may be even better than brush modification. Adding just new faces would mean not creating redundant inner geometry and only focusing on the outer, visible structure.

There's a few places that it could go, but I'm leaning towards `generate_brush_vertices` to avoid messing with the map parser directly.

Going to try to get a prototype together tonight, will let you know if I can get it working.

Ralane: After playing around with a few different approaches, I don't think there's a great way to do this with func_godot, at least not without making a mess of the rest of the core functionality.

I tried rebuilding the vertex array with more vertices, and while I was able to calculate the new locations pretty easily, the problem is that `wind()` sorts by angles so it won't support more complex shapes. So the other approach might be combining that approach with either adding more faces, or adding more brushes.

I did also find some libraries that handle mesh subdivisions, so in theory you could run those against the generated mesh, apply some noise, and then build a trimesh collider out of it. 

But at that point it'd be easier to use a different tool entirely. Subdividing in the map editor might still just be the easiest approach, or using a 3d model for the more detailed areas.

PR #276, opened 2026-09-13
Exclude entirely Skip brushes from Convex collider generation

Ralane:
Exclude entirely Skip brushes from Convex collider generation. Allows users to use Skip brushes as bounding boxes, to provide snapping/AABBs around otherwise uneven geometry.

As an example, consider the following pipe props. Due to the uneven shape and decorative rings, aligning them with snapping would otherwise be difficult. This change allows us to instead use Skip to align them along the grid modularly, without impacting the generated map.

image

image

### DIFF
--- a/addons/func_godot/src/core/geometry_generator.gd
+++ b/addons/func_godot/src/core/geometry_generator.gd
@@ -37,6 +37,13 @@ func _init(settings: FuncGodotMapSettings = null, hplane_size: float = 512.0) ->
 func is_skip(face: _FaceData) -> bool:
 	return FuncGodotUtil.is_skip(face.texture, map_settings)
 
+# One niche use case for Skip brushes is providing a bounding box for clean snapping/AABBs to an otherwise uneven convex prop
+func is_brush_entirely_skip(brush: _BrushData) -> bool:
+	for face in brush.faces:
+		if(!is_skip(face)):
+			return false;
+	return true;
+
 func is_clip(face: _FaceData) -> bool:
 	return FuncGodotUtil.is_clip(face.texture, map_settings)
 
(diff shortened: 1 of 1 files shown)


RisingThumb:
Can you show if this actually affects the AABB of the generated brush? I don't think it would? It'd certainly change the origin point of your brush entity though because the skip brush contributes to that I think? If it's entirely skip, it shouldn't generate a mesh as well for that brush so the AABB should be exactly as a big as the pipe I thought?

Since you mentioned alignment, consider using an origin brush to set the point of origin for easy alignment. Maybe I misunderstand your alignment issue.

This change potentially causes issues with current projects if anyone uses skip brushes for trigger volumes (Area3Ds) instead of clip brushes.

Ralane:
I didn't realize that origin brushes were a thing 🤦 seems I've reinvented the wheel here and should have been using those instead since they functionally do the same thing.

Ralane:
Well, not exactly the same thing, since origin brushes also set the origin, but at least the exclusion from convex geometry aspect

RisingThumb:
@Ralane I'd still keep the PR open. I think brushes that are full skip should skip convex collisions, but my comments were more directed to your problem and not the change itself.

Another thing to consider is that your different solid classes have different entity origin settings for the origin of a brush entity. I'd recommend checking those too, but of course the origin brush gives you fine control

RhapsodyInGeek:
I'm really not a fan of the added rule complexity to an already complex set of rules, tbh. We've already carved out an exception for Skip faces in terms of concave vs convex, so to carve out another exception in regards to convex vs convex seems a bit much.

I'm also hesitant to change build rules because this kind of change could break other users' projects. What if someone relies on a full skip brush in their own entities for collision purposes? Did we just break their project? Are we going to add yet another optional property to the SolidClass resource in order to maintain backwards compatibility with the current expected build rules?

If this is solely for grid alignment on the TB side, you could honestly do it by grouping the alignment brush with the pipe into a TB Group (not an entity) and then either:
- leave that alignment brush as worldspawn if you don't build worldspawn
- move that brush to an Alignment Layer that is set to Omit, which will skip it entirely
In either case, you wouldn't bind the alignment brush to the same entity as the pipe brush entity.

As it is, I don't see a large enough benefit to merge this PR considering the added rules complexity and backwards compatibility concerns, especially given that there are convenient enough alternatives.

